PR #10, opened 2023-09-27
FRONT-712: Extend ErrorContent with IllustrationPlaceholder's size & direction

DC-RomanKarpov:
Remove some unused styles

### DIFF
--- a/src/ui/components/ErrorContent/ErrorContent.js
+++ b/src/ui/components/ErrorContent/ErrorContent.js
@@ -43,16 +43,14 @@ class ErrorContent extends React.PureComponent {
         noControls: PropTypes.bool,
         showDebugInfo: PropTypes.bool,
         noActions: PropTypes.bool,
-        size: PropTypes.oneOf('s', 'm', 'l', 'promo'),
-        direction: PropTypes.oneOf('row', 'column'),
+        size: PropTypes.oneOf(['s', 'm', 'l', 'promo']),
+        direction: PropTypes.oneOf(['row', 'column']),
     };
 
     static defaultProps = {
         type: ErrorContentTypes.ERROR,
         showDebugInfo: true,
         noActions: false,
-        size: 'l',
-        direction: 'row',
     };
 
     state = {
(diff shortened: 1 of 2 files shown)


src/ui/components/ErrorContent/ErrorContent.scss
```diff
@@ -32,13 +32,24 @@
...
+    }
+
     &_no-controls {
         display: block;
         padding: 20px 0;
     }
 
     &__title {
-        font-weight: 500;
+        font-weight: var(--g-text-header-font-weight);
```
ariser, line 52: we have this variable set to 500 in DC code
but isn't this is a component that DL also uses?
I don't know if they have the same override

ariser (commented):
lgtm, but I suppose we should also show this to someone from DL team

src/ui/components/ErrorContent/ErrorContent.scss
DC-RomanKarpov: Good point. Will remove. Thanks

mournfulCoroner (approved):

ariser (approved):

src/ui/components/ErrorContent/ErrorContent.js
```diff
@@ -272,12 +276,19 @@ class ErrorContent extends React.PureComponent {
...
+                    className,
+                )}
+            >
                 <PlaceholderIllustration
                     name={imageName}
                     title={this.renderTitle()}
                     description={this.renderDescription()}
                     renderAction={() => this.renderActions()}
+                    size={size}
+                    direction={direction}
```
mournfulCoroner, line 291: By default in `PlaceholderIllustration` direction switches to column in mobile view. This behavior may get lost if we pass direction in that way. I think you need to check here by DL.IS_MOBILE too

mournfulCoroner (changes requested):

src/ui/components/ErrorContent/ErrorContent.js
DC-RomanKarpov: Thank you for the remark. Will add this behavior.

mournfulCoroner (approved):

ariser (changes requested):

ariser (approved):

PR #36, opened 2023-10-05
New relations dialog improvements

Marginy605:


src/ui/units/dash/containers/Dialogs/DialogRelations/types.ts
```diff
@@ -55,6 +57,8 @@ export type ConnectionsData = Array<{
 export type AliasClickHandlerData = {
     showDebugInfo: boolean;
     currentRow: DashkitMetaDataItem;
+    relationText: React.ReactNode | string;
```
mournfulCoroner, line 60: Doesn't ReactNode include string?

src/ui/units/dash/containers/Dialogs/DialogRelations/components/Content/Row.tsx
```diff
@@ -158,34 +160,62 @@ export const getTooltipInfo = ({
...
             : undefined;
 
-    return {tooltipContent, tooltipTitle};
+    const aliasDetailTitle = (
+        <React.Fragment>
+            {getRelationDetailsText({
+                text: i18n(getRelationDetailsKey(relationType)),
+                linkType: relationType as RelationType,
+                widget: `${widgetLabel}${widgetSeparator}${widgetTitle}`,
+                row: `${rowLabel}${rowSeparator}${rowTitle}`,
```
mournfulCoroner, line 207: Row and widget are identical here and in `tooltipTitle`. Perhaps we should put this in a variable?

src/ui/units/dash/containers/Dialogs/DialogRelations/types.ts
Marginy605: According to @types react it doesn't.
image

src/ui/units/dash/containers/Dialogs/DialogRelations/components/Content/Row.tsx
Marginy605: It is not final variant of texts, I'm sure after design review we will change it not once

src/ui/units/dash/containers/Dialogs/DialogRelations/types.ts
mournfulCoroner: Well hidden

Marginy605: fixed

mournfulCoroner (approved):

PR #57, opened 2023-10-12
Aliases popup add logic, fix styles

Marginy605:
- Add alias add widget icons, fix styles
- Add alias add validation
- Add save new added alias logic
- Fix update aliases bug (on remove last alias)
- Add filters for alias options list (in add new alias block)


Marginy605:
Texts will be changed later (before enabling feature flag with new relations)

mournfulCoroner (approved):

PR #82, opened 2023-10-17
Fixed changed relation type, aliases update bugs

Marginy605:


mournfulCoroner (approved):

src/shared/schema/mix/actions/dash.ts
```diff
@@ -41,49 +41,76 @@ export const dashActions = {
...
-                            datasetId,
-                            version: 'draft',
-                        });
+        const fetchEntries = entries.map(async (entry) => {
+            const {entryId, type, meta, links} = entry;
+            const {dataset} = links || {};
+            // deprecated
+            const {datasetId: metaDatasetId} = meta || {};
+            const datasetId = (dataset || metaDatasetId) as string | undefined;
+            const widgetType = type.match(/^[^_]*/)?.[0] || null;
```
goshander, line 50: Maybe `at least one` regex with `+`  sign
image

goshander (approved):

src/shared/schema/mix/actions/dash.ts
Marginy605: If you check full expression, you will see no changes in result:
image

PR #84, opened 2023-10-18
Fix unexistent columns in visualization

artemipanchuk:


src/server/modes/charts/plugins/ql/js.ts
```diff
@@ -188,10 +195,22 @@ export default ({shared, ChartEditor}: {shared: QLEntryDataShared; ChartEditor:
...
+                fields,
+                items: sharedLabels as Field[],
+            });
+
+            newShapes = mapItems({
+                fields,
+                items: sharedShapes as Field[],
+            });
+
+            console.log(newColors, newLabels, newShapes);
```
kuzmadom, line 213: console.log should probably be deleted

src/server/modes/charts/plugins/ql/utils/visualization-utils.ts
```diff
@@ -191,8 +193,10 @@ const mapItems = ({items, fields}: {items: Field[]; fields: Field[]}) => {
             }
         }
 
-        // TODO: else we need to make this field invalid
-        return item;
+        return {
+            ...item,
+            conflict: 'not-existing-ql',
```
kuzmadom, line 198: if there is some kind of enum or constant, it is better to use it instead of a string

PR #85, opened 2023-10-18
New aliases bugfixes

Marginy605:
- Fix update relations list types after aliases applying (if there are other relations which could be connected by the same alias)
- Add validation of the same fields on aliases add
- Fix update relations list after add alias in one row and delete in other


mournfulCoroner (approved):

Marginy605:
Random tests failed, not connected with code in PR

PR #96, opened 2023-10-23
Add icons to workbook collection actions

Sergey-weber:
Added icons to workbook/collection actions in CollectionContentTable.tsx


src/ui/units/collections/components/CollectionContentTable/CollectionContentTable.tsx
```diff
@@ -61,6 +62,39 @@ type Props = {
     refreshContent: () => void;
 };
 
+const dropdownActions: {[key: string]: React.ReactNode} = {
+    move: (
+        <div className={b('dropdown-item')}>
+            <Icon data={ArrowRight} />
+            <div className={b('dropdown-text')}>{i18n('action_move')}</div>
```
imsitnikov, line 69: 1) We shouldn't use `i18n` from non-react context, `i18n` might be uninitialized in this context. We need to move this into a react component and use `useMemo`.
2) Let's store only data, without tags copying:
```typescript
const dropdownActions = useMemo(() => {
    return {
        move: {
            title: i18n('action_move'),
            icon: ArrowRight,
        },
        ...
    }
}, []);
```

imsitnikov (changes requested):

jhoncool:
@Sergey-weber can u text some desciption?

src/ui/units/collections/components/CollectionContentTable/CollectionContentTable.tsx
Sergey-weber: yup, maybe better to create new component in other file and add props for setting - icon component and text?

imsitnikov (approved):

jhoncool (approved):

PR #129, opened 2023-10-27
Mock data for E2E tests

bt4R9:
In this PR
- New database `sales` running on existing `postgres-connection`
- Mock data is taken from the open source
- Connection and Dataset that use the new database for US config
- e2e-data-volume to share data to the US container

Related PR in US https://github.com/datalens-tech/datalens-us/pull/21


tests/docker-compose.override.yml
```diff
@@ -1,8 +1,24 @@
 version: "3.7"
 
 services:
+  us:
+    build:
+      context: ../datalens-us
+      dockerfile: ../datalens-us/Dockerfile
```
vrozaev, line 7: TODO: remove the build completely. We need it now only for debug during development.

vrozaev (approved):

PR #158, opened 2023-11-01
Disable default sorting for ql charts

artemipanchuk:


kuzmadom (changes requested):
The problem with the commented `order by`

src/ui/units/wizard/components/Dialogs/DialogPlaceholder/DialogPlaceholder.tsx
```diff
@@ -391,30 +391,57 @@ class DialogPlaceholder extends React.PureComponent<Props, State> {
...
+        let disabledTooltipContent = null;
+
         const radioButtons = AXIS_MODE_RADIO_BUTTONS.map((option) => {
             if (option.value === SETTINGS.AXIS_MODE.CONTINUOUS) {
+                // Axis mode always depends on first item in placeholder
+                let disabled = !isAllAxisModesAvailable(firstField) || sortHasMeasures;
+
+                if (disabled) {
+                    showDisabledTooltip = true;
+                    disabledTooltipContent = 'label_axis-mode-unavailable-type';
```
kuzmadom, line 404: I think you need to either write the more generalized text in the tooltip, or separate the keys for different disable cases.
Now, in the case of sorting, there will be a text that does not explain the reason for the disabling.

src/server/modes/charts/plugins/datalens/preparers/bar-x/d3.ts
```diff
@@ -23,14 +23,17 @@ type OldBarXDataItem = {
 } | null;
 
 export function prepareD3BarX(args: PrepareFunctionArgs): ChartKitWidgetData {
-    const {shared, labels, placeholders} = args;
+    const {shared, labels, placeholders, applyDefaultSorting = true} = args;
```
kuzmadom, line 26: Below, the default sorting is applied if there are no fields in the corresponding placeholder. It seems to me that in your case it would be more appropriate to use a flag to disable sorting. Something like `disableSorting` maybe?

src/server/modes/charts/plugins/datalens/preparers/pie/__tests__/pie.test.ts
```diff
@@ -36,8 +36,8 @@ describe('preparePie', () => {
         }));
 
         expect(items).toEqual([
-            {color: '#4DA2F1', colorValue: '2'},
-            {color: '#FF3D64', colorValue: '1'},
+            {color: '#4DA2F1', colorValue: '1'},
```
kuzmadom, line 39: Why do you need to edit tests? They should have kept the old behavior, shouldn't they?

src/server/modes/charts/plugins/ql/js.ts
```diff
@@ -246,6 +252,24 @@ export default ({shared, ChartEditor}: {shared: QlConfig; ChartEditor: IChartEdi
             });
         }
 
+        if (newVisualization.placeholders[0]?.items[0]) {
```
kuzmadom, line 255: It is better to refer the placeholder by id, not by index

src/server/modes/charts/plugins/ql/js.ts
```diff
@@ -246,6 +252,24 @@ export default ({shared, ChartEditor}: {shared: QlConfig; ChartEditor: IChartEdi
...
+                (LINEAR_VISUALIZATIONS.has(newVisualization.id) ||
+                    newVisualization.id === WizardVisualizationId.BarXD3)
+            ) {
+                newVisualization.placeholders[0].settings = {
+                    axisModeMap: {
+                        [newVisualization.placeholders[0].items[0].guid]: 'discrete',
+                    },
+                };
+
+                newVisualization.placeholders[0].items[0].disableAxisMode = true;
```
kuzmadom, line 267: `disableAxisMode`
Why is this a characteristic of the field? Perhaps you should specify this in the X-axis placeholder settings?

src/server/modes/charts/plugins/ql/js.ts
```diff
@@ -246,6 +252,24 @@ export default ({shared, ChartEditor}: {shared: QlConfig; ChartEditor: IChartEdi
...
+            } else {
```
kuzmadom, line 268: But in general, I don't think it is necessary to store a flag for the field disabling. It would be more useful to keep the knowledge about the sorting in the query (may be somewhere in metadata?). In this case, the field disable flag can be calculated, and the knowledge about sorting can be used somewhere else.

src/server/modes/charts/plugins/ql/js.ts
```diff
@@ -246,6 +260,30 @@ export default ({shared, ChartEditor}: {shared: QlConfig; ChartEditor: IChartEdi
             });
         }
 
+        if (visualizationCanHaveContinuousAxis(newVisualization)) {
+            const targetPlaceholderId = HORIZONTAL_VISUALIZATIONS.has(newVisualization.id)
+                ? 'y'
+                : 'x';
```
kuzmadom, line 266: Need to use enum `PlaceholderId`

src/shared/types/config/wizard/v9.ts
```diff
@@ -22,7 +22,7 @@ export type V9ChartsConfig = {
     links: V9Link[];
     sort: V9Sort[];
     tooltips: V9Tooltip[];
-    type: 'datalens';
+    type: 'datalens' | 'ql';
```
kuzmadom, line 25: It is better to delete 'ql'

kuzmadom (approved):

PR #170, opened 2023-11-03
Add support chart-chart filtering in wizard

kuzmadom:


korvin89 (approved):

src/ui/libs/DatalensChartkit/ChartKit/helpers/apply-hc-handlers.ts
```diff
@@ -234,6 +30,19 @@ export const fixPieTotals = (args: {data: GraphWidget}) => {
     });
 };
 
+function getActionParams(params: Record<string, string | string[]> = {}) {
```
Marginy605, line 33: We have already functions in dashkit for getting actionParams from params, see `pickActionParamsFromParams`. And it has different formats to return (with prefix or not).

src/ui/components/Widgets/Chart/Chart.tsx
```diff
@@ -170,7 +170,7 @@ export const Chart = (props: ChartNoWidgetProps) => {
         ignoreUsedParams,
         clearedOuterParams,
         onInnerParamsChanged,
-        enableActionParams: true,
+        enableActionParams: false,
```
Marginy605, line 173: For this change we need to add new reserved get-parameter to enable it in preview. And add it;
- for forming link to open in new tab (and remove from link for edit, as far as I remember, there is single helper for both)
- support force adding get-parameter in mailings

Marginy605 (changes requested):

src/ui/components/Widgets/Chart/Chart.tsx
kuzmadom: Added a separate parameter `_ap_enabled` (used for preview). The link generation for screenshots will be done in a separate task.

src/ui/constants/common.ts
```diff
@@ -255,6 +255,7 @@ export const URL_OPTIONS = {
     EMBEDDED: '_embedded',
     NO_CONTROLS: '_no_controls',
     LANGUAGE: '_lang',
+    ACTION_PARAMS_ENABLED: '_ap_enabled',
```
Marginy605, line 258: Bad name - dashkit will think that there is action param `enabled` for this chart

Marginy605 (approved):

PR #176, opened 2023-11-07
Fix colors when using mounted colors and auto colors

artemipanchuk:


Flunt1k:
As a suggestion: We already have a list of colors used (`colorsConfig.mountedColors`), it looks like we can avoid the first pass through the array and do the coloring in one pass through the graphs. 
I suppose we could do something like `Object.values(colorsConfig.mountedColors)` and then add that to `usedColors`. After that, we can make a single pass through the graphs and color them

Flunt1k (approved):

PR #177, opened 2023-11-07
Add and use redux/toolkit

imechoim:


src/ui/units/dash/containers/Dialogs/Settings/Settings.tsx
```diff
@@ -125,6 +126,7 @@ const Settings = () => {
                     loadPriority,
                 }),
             );
+            dispatch(parametersGlobalSlice.actions.update(globalParams));
```
korvin89, line 129: U can wrap all dispatches to one [batch](https://react-redux.js.org/api/batch)

PR #178, opened 2023-11-07
Fix legend visibility for legacy monitoring charts

artemipanchuk:


Flunt1k (approved):

src/server/modes/charts/plugins/ql/yagr.ts
```diff
@@ -68,7 +68,8 @@ export default ({shared}: {shared: QlConfig}) => {
     });
 
     const isLegendEnabled = Boolean(
-        config.colors?.length && config.extraSettings?.legendMode !== 'hide',
+        (config.colors?.length && config.extraSettings?.legendMode !== 'hide') ||
+            config.extraSettings?.legendMode !== 'show',
```
Flunt1k, line 72: It looks like the mistake in the condition. isLegendEnabled and `legendMode !== 'show'` seems controversial

Flunt1k (approved):

PR #184, opened 2023-11-13
Add loading dash entries datasets

Marginy605:
- Collect dash chart ids and selectors datasets ids, then loading datasets schemas for them
- Add dataset fields in params for wizard charts to dashkit config; clear them before setting updated config


Marginy605:
There are several ts ignores in code because of problems of forming full object step by step (by partly filling fields).
If you have good variants for solving this problems, please welcome.
The original ts error looks like that:
image

mournfulCoroner (approved):

PR #209, opened 2023-11-16
Color mode change support

artemipanchuk:


src/server/modes/charts/plugins/datalens/preparers/bar-x/prepareBarX.ts
```diff
@@ -453,13 +456,13 @@ export function prepareBarX(args: PrepareFunctionArgs) {
             });
         });
 
-        if (isColorizeByMeasure || isColorizeByMeasureValue) {
-            colorizeByMeasure(visualizationId as WizardVisualizationId, {
+        if (colorMode === ColorMode.GRADIENT || isColorizeByMeasure || isColorizeByMeasureValue) {
```
kuzmadom, line 459: It is better to put it in a separate utility (duplicated in several more places by code)

src/server/modes/charts/plugins/datalens/preparers/line/helpers/line-mappers/mapDataToDimensionColoredLines.ts
```diff
@@ -164,13 +167,22 @@ export const mapDataToDimensionColoredLines = ({
...
     const lastKey = typeof xValue === 'undefined' ? shownTitle : xValue;
 
-    lines[key].data[lastKey as string | number] = {value: yValue};
+    const pointData: {
+        value: typeof yValue;
+        colorValue?: number;
+    } = {value: yValue};
+
+    if (colorModeIsGradient) {
+        pointData.colorValue = Number(lines[key].colorValue);
```
kuzmadom, line 182: There is already a cast to Number above. Do you need to duplicate here?

src/server/modes/charts/plugins/datalens/preparers/polyline.ts
```diff
@@ -162,10 +162,10 @@ const preparePolyline = (options: PrepareFunctionArgs) => {
             {},
         );
 
-        if (color.type === DatasetFieldType.Measure) {
-            colorData = colorizeGeoByMeasure(hashTable, colorsConfig).colorData;
+        if (isMeasureField(color)) {
```
kuzmadom, line 165: There is no need to take into account _colorMode_?

src/ui/units/wizard/components/Dialogs/DialogColor/DialogColor.tsx
```diff
@@ -54,21 +55,36 @@ export type OpenDialogColorArgs = {
...
+class DialogColorComponent extends React.Component<Props, State> {
+    constructor(props: Props) {
+        super(props);
+
+        // isGradient is legacy fallback flag
+        // colorMode is modern value
+        this.state = {
+            colorMode:
+                props.colorsConfig?.colorMode ||
+                (props.isGradient ? ColorMode.GRADIENT : ColorMode.PALETTE),
```
kuzmadom, line 71: Fields that could have been preserved in the old features need to be mapped through the config migration mechanism.

src/ui/units/wizard/selectors/dialogColor.ts
```diff
@@ -55,3 +64,26 @@ export function isGradientDialog(
...
+
+export function isColorModeChangeAvailable(
+    props: Pick<OpenDialogColorArgs['props'], 'item'> & {
+        visualizationId?: string;
+    },
+) {
+    const {item, visualizationId} = props;
+
+    const itemIsSuitable =
+        item.type === DatasetFieldType.Dimension && item.data_type === DATASET_FIELD_TYPES.INTEGER;
```
kuzmadom, line 82: is the float type not considered suitable for a gradient?

artemipanchuk: According to Pavel's plan - yes

src/ui/units/wizard/components/Dialogs/DialogColor/DialogColor.tsx
artemipanchuk: If we want to do it that way, we should add v10 version, because now acutal version is v9.

src/ui/units/wizard/selectors/dialogColor.ts
artemipanchuk: But I think it will be ok

src/server/modes/charts/plugins/datalens/preparers/line/helpers/line-mappers/mapDataToDimensionColoredLines.ts
artemipanchuk: We cast `line.colorValue` above. And here we cast `point.colorValue`.

kuzmadom (approved):

PR #215, opened 2023-11-20
Add waiting ds fields & load displaying for chart-by-chart filtering case

Marginy605:
- Add logic for waiting dataset fields for wizard filtering charts if we triggered changed before getting dataset fields schema
- Split api to `getEntriesDatasetsFields` for relations (as it had been before changes) and `getWidgetsDatasetsFields` for filtering chart-by-chart logic


src/ui/components/Widgets/Chart/ChartWidget.tsx
```diff
@@ -79,13 +87,28 @@ export const ChartWidget = (props: ChartWidgetProps) => {
...
+                | null,
+        [chartId, dashDatasetFields],
+    );
+    const prevWidgetDatasetFields = usePrevious(widgetDatasetFields);
+
+    const loadedDSFields = widgetDatasetFields && widgetDatasetFields !== prevWidgetDatasetFields;
+
     const chartkitParams = React.useMemo(() => {
-        let res = removeEmptyProperties(props.params);
+        let res = removeEmptyNDatasetFieldsProperties(
```
Marginy605, line 102: We need to reject extra re-render (and api re-run) because of changed dataset fields params after loading dataset fields on dash

PR #219, opened 2023-11-21
Change action params api for Table plugin

korvin89:
[CHARTS-8726]


src/ui/libs/DatalensChartkit/ChartKit/components/Widget/components/Table/utils/misc.ts
```diff
@@ -342,3 +336,182 @@ export function getTreeSetColumnSortAscending(
...
+    const rowActionParams = getRowActionParams(row);
+    const isRowAlreadySelected = isRowSelected({actionParams, rowActionParams});
+
+    if (isRowAlreadySelected) {
+        Object.keys(rowActionParams).forEach((key) => {
+            rowActionParams[key] = [''];
+        });
+    }
+
+    const resultParams = {} as Record<string, string | string[]>;
```
kuzmadom, line 450: `StringParams` ?

Marginy605:
Before changes you could set for cell only one event `setParams` OR `setActionParams`. This is base behaviour, because otherwise we cannot guarantee consistency.  As far as I understand, there is possible situation when I can set row event `setActionParams` AND cell event `setParams`.

kuzmadom (approved):

PR #223, opened 2023-11-21
Add query name value editing for QL chart

Flunt1k:
- Fix yarn tooltip bug
There was inappropriate usage of `tracking` option. It was equal to `visualizationId` or `highchartsId`, it leaded to lose user's cursor tracking when hovering over lines.  According to the type, this could only be a `'sticky'|'area'|function` value. 
- Add mapping config to version 2
QL config now supports a required queryName field. By default, the `Query [index]` construct is used.
- Add editable text component
Add component that allows to edit text value on click via replacing `div` to `Input`

### DIFF
--- a/src/server/modes/charts/plugins/ql/yagr.ts
+++ b/src/server/modes/charts/plugins/ql/yagr.ts
@@ -1,6 +1,6 @@
 import type {ChartType, YagrWidgetData} from '@gravity-ui/chartkit/yagr';
 
-import {IChartEditor, ServerVisualization} from '../../../../../shared';
+import {IChartEditor, QlVisualizationId, ServerVisualization} from '../../../../../shared';
 import {mapQlConfigToLatestVersion} from '../../../../../shared/modules/config/ql';
 import type {QlConfig} from '../../../../../shared/types/config/ql';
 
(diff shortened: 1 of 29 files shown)


src/server/modes/charts/plugins/ql/yagr.ts
```diff
@@ -43,16 +43,17 @@ const applyPlaceholderSettingsToYAxis = ({
...
     const type = (config.visualization.highchartsId || config.visualization.id) as ChartType;
 
     const percent =
         config.visualization.id === 'area100p' || config.visualization.id === 'column100p';
 
-    const tracking = (config.visualization.highchartsId ||
-        config.visualization.id) as TooltipOptions['tracking'];
+    const visualizationId = config.visualization.highchartsId || config.visualization.id;
+
+    const tracking = visualizationId === 'area' ? 'area' : 'sticky';
```
artemipanchuk, line 56: Minor: I think we have an enum for 'area'

Flunt1k: I don't think this is quite right. The fact that the type from Yagr matches our visualizationId is a coincidence rather than a pattern.

My suggestion is to leave it as it is, so as not to mix up our `area` and what Yagr means by it.

kuzmadom (approved):

artemipanchuk (approved):

PR #240, opened 2023-11-23
Add extending API

melikhov-dev:


jhoncool (approved):

api/server/components.ts
```diff
@@ -0,0 +1,20 @@
...
+export {
+    getLandingLayout,
+    SDK,
+    Utils,
+    getChartkitLayoutSettings,
+    getPlatform,
+} from '../../src/server/components';
+export {getAppLayoutSettings} from '../../src/server/components/app-layout/app-layout-settings';
+export * as resolveEntryByLink from '../../src/server/components/resolve-entry-by-link';
+export * as metrikaDataFormatter from '../../src/server/components/metrika-data-formatter';
```
imsitnikov, line 11: Would be better not to use the `*` exports

imsitnikov (approved):

api/server/components.ts
melikhov-dev: Yep, I removed them

PR #241, opened 2023-11-23
Add auto run monitoring query on change

Flunt1k:
Fiche allows the user to not have to click the Run button to perform Graph Monitoring. 
Entering query and pressing cmnd/ctrl + enter will execute the chart. 
Also if focus is lost in the text area, the chart will be executed. 

With this feature turned off, you need to press the Run button each time.


src/shared/types/wizard/index.ts
```diff
@@ -94,6 +94,7 @@ export interface CommonSharedExtraSettings {
     navigatorSettings?: NavigatorSettings;
     enableGPTInsights?: boolean;
     labelsPosition?: LabelsPositions;
+    qlAutoExecuteChart?: 'on' | 'off';
```
kuzmadom, line 97: Perhaps it is worth dividing the parameters into Wizard and QL ? To have something like `WizardChartSetting | QlChartSetting`.

In this case, you will not need to use the ql prefix.

src/i18n-keysets/sql/ru.json
```diff
@@ -52,5 +52,6 @@
   "label_visualization-pie": "Круговая диаграмма",
   "label_visualization-scatter": "Точечная диаграмма",
   "label_visualization-treemap": "Древовидная диаграмма",
-  "text_default-name": "Новый QL-чарт"
+  "text_default-name": "Новый QL-чарт",
+  "label_ql-auto-execution-chart": "Автоматическое выполнение при изменении"
```
kuzmadom, line 56: Автоматически применять изменения

kuzmadom (approved):

src/shared/types/wizard/index.ts
Flunt1k: Let's do the refactoring in the next PR

PR #244, opened 2023-11-24
Move QL tests

imechoim:


vrozaev (approved):

vrozaev (approved):

tests/constants/test-entities/connections.ts
```diff
@@ -0,0 +1,7 @@
+export enum ConnectionsUrls {
+    Connection = '/connections/dr0y3yyi1gjk2-connection',
```
Flunt1k, line 2: I suggest to give more more meaningful names for entries. Because there can be more than one connection and it would be nice to determine them by name

vrozaev: +1

vrozaev (changes requested):

tests/page-objects/dashboard/DashboardPage.ts
```diff
@@ -641,6 +641,14 @@ class DashboardPage extends BasePage {
     async isDescriptionEditMode() {
         await this.description.isEditMode();
     }
+
+    async clickSelectWithTitle(title: string) {
+        await this.page
+            .locator(slct('chartkit-control'))
+            .filter({hasText: title})
+            .filter({has: this.page.locator(slct('chartkit-control-select'))})
```
Marginy605, line 649: `chartkit-control` & `chartkit-control-select` must be constants, and may be it would better to get it as function arguments (now it looks like hardcode)

Marginy605 (changes requested):

tests/page-objects/dashboard/DashboardPage.ts
imechoim: Updated PO. Added the use of constants. I think it should be a precision selector. Passing `data-qa` arguments will only make it harder to use.

Marginy605 (approved):

PR #257, opened 2023-11-27
Move dash loading tests

imechoim:


tests/page-objects/dashboard/DashboardPage.ts
```diff
@@ -85,6 +85,8 @@ class DashboardPage extends BasePage {
         acceptableSelectBtn: 'select-acceptable-button',
         dialogApplyBtn: 'dialog-apply-button',
         dialogCancelBtn: 'dialog-cancel-button',
+        chartGridItemContainer: `${slct(COMMON_DASH_SELECTORS.DASH_GRID_ITEM)} .chartkit`,
+        dashPluginWidgetBody: slct('chart-widget'),
```
mournfulCoroner, line 89: I'm not sure it's a good idea to change api (adding `slct()`) for `selectors` object only for two new values. This will be confusing for future usage

imechoim: It's just selectors. It's preferable to put selectors in one location rather than having many files for constants.

imechoim: The `slct()` creates a wrapper to use the `data-qa` attribute. As a result you get just a selector.

mournfulCoroner (approved):

Marginy605 (approved):

PR #268, opened 2023-11-28
Move dash tabs tests

imechoim:


Marginy605 (approved):

goshander:
@imechoim this type throw exception on closed part if was been extended in open part
https://github.com/datalens-tech/datalens-ui/blob/14735e8562f37d6e1c820460084559f6f3b17cb7/tests/types/config/index.ts#L6

it is expected behaviour?

imechoim:
Yes. This is expected behaviour. Just add new ids in closed part.

PR #295, opened 2023-12-01
Add checkbox dash selector type

mournfulCoroner:


src/server/modes/charts/plugins/control/js/index.ts
```diff
@@ -174,6 +174,19 @@ export default ({
 
             break;
 
+        case ELEMENT_TYPE.CHECKBOX:
+            uiControl.type = source.elementType;
+
+            if (!params[param]) {
+                if (typeof source.defaultValue === 'undefined') {
```
goshander, line 181: Maybe strict compare with `=== undefined` is better here?

PR #310, opened 2023-12-05
Fixed connection choose for ql charts: now we can choose any type of connection (because it's could be any type)

vrozaev:
**Expected behaviour**:
User can create QL chart for each connection which is allowed to do so. 

User can create QL chart from connection page or from new ql chart page `/ql`.

**Actual behaviour**:
User can create QL chart for each connection which is allowed to do so. 

User can create QL from connection page **but he can't** create some type of QL charts new ql chart page `/ql` because some connections maked as "inactive".

image

This pull request introduce the **expected behaviour**:
image


vrozaev:
Fix is incorrect, the correct fix is enabling QLPrometheus feature.

PR #316, opened 2023-12-07
Add mobile version of chart menu

mournfulCoroner:


src/ui/libs/DatalensChartkit/components/ChartKitBase/components/Header/components/Menu/Menu.js
```diff
@@ -199,29 +223,57 @@ export class Menu extends React.PureComponent {
         }
 
         const menuItems = this.prepareItems(visibleItems, data, onChange);
-        const ModalComponent = this.state?.modal;
+
+        const mobileItems = menuItems.flat();
```
goshander, line 227: Can we traenform menuItems to flat array only if IS_MOBILE

goshander (approved):

src/ui/libs/DatalensChartkit/components/ChartKitBase/components/Header/components/Menu/Menu.js
mournfulCoroner: `menuItems.flat()` is moved inside the block with the `renderSheet`

PR #324, opened 2023-12-08
Refactor unit tests, added basic ql unit tests for all visualizations

artemipanchuk:
1) Renamed some unit tests
2) Removed unneccessary parts of mocks for some unit tests
3) Added new unit tests for basic ql charts, same as in smoke tests
4) Refactored some enums in prepare-single-result
5) Refactored some types in ql plugin (added v3)


src/ui/units/ql/store/utils/grid.ts
```diff
@@ -105,7 +105,11 @@ export class Helper {
                 entryId: '',
                 type: '',
             },
-            version: QlConfigVersions.V2,
+            colors: [],
+            shapes: [],
+            labels: [],
+            tooltips: [],
```
kuzmadom, line 111: Can there be tooltips in QL too?

src/shared/types/config/ql/v3.ts
```diff
@@ -0,0 +1,101 @@
...
+    capacity?: number;
+    size: number;
+    id?: string;
+    allowedTypes?: string[];
+    pseudo?: boolean;
+}
+
+export interface QLResultEntryMetadataDataColumnV3 {
+    // typeName -- legacy type that is incompatible with Field
+    // How to turn on wizard ql common visualization -- you can delete
```
kuzmadom, line 18: ?

kuzmadom (approved):

src/ui/units/ql/store/utils/grid.ts
artemipanchuk: Currently no, but I think we should have them for compatibility and probably we will have to add support of geovisualizations and tooltips in future

src/shared/types/config/ql/v3.ts
artemipanchuk: Fixed

PR #328, opened 2023-12-11
Add mobile version of dash TOC 

mournfulCoroner:


src/ui/units/dash/containers/Header/Header.tsx
```diff
@@ -34,28 +42,34 @@ type Props = StateProps & DispatchProps & OwnProps;
...
-            history: this.props.history,
-        });
-    };
-
     render() {
         if (!this.props.entry) {
             return null;
         }
 
+        const handleTocHeaderClick = () => this.props.toggleTableOfContent();
```
Marginy605, line 62: Optional: there is no need to use extra function for one line function. Could you use 
`<MobileTocHeader onClick={this.props.toggleTableOfContent} />` here?

PR #339, opened 2023-12-14
Fix request for datasets without edit rights in relations 

mournfulCoroner:


src/shared/schema/mix/actions/dash.ts
```diff
@@ -60,7 +60,22 @@ export const dashActions = {
         );
 
         const allDatasetsFetchedData = await Promise.all([...allDatasetsPromises]);
-        const allDatasetsFetchedDataDict = allDatasetsFetchedData
+
+        // for requests that ended in error, we try to get the dataset fields using `fetchDatasetFieldsById`, which doesn't require edit rights to the dataset
```
Marginy605, line 64: or requests that ended **with** error

Marginy605 (changes requested):

Marginy605 (approved):

PR #348, opened 2023-12-15
Column direction for the Chart preview section

bt4R9:
This PR changes the illustration/text direction from the "row" to "column" to ensure consistent style throughout the service.

Before
Screenshot 2023-12-15 at 15 22 48

After
Screenshot 2023-12-15 at 15 22 26


bt4R9:
@melikhov-dev It seems like. IIRC the decision was to investigate the issue in the whole UI by the Design team and then create a PR to fix this issue in all places.

PR #361, opened 2023-12-19
Fix rerender widget with empty params on clearing

Marginy605:


src/ui/components/Widgets/Chart/hooks/useLoadingChart.ts
```diff
@@ -211,13 +211,23 @@ export const useLoadingChart = (props: LoadingChartHookProps) => {
...
-                },
-            };
+            // when clear params of widget from outer (ex cleared actionParams from other chart cause empty params in curent)
+            if (hasChangedOuterParams && isEmpty(params)) {
+                res = {
+                    ...res,
+                    params: {
+                        ...params,
+                    },
+                };
```
mournfulCoroner, line 221: Maybe here and in the other two cases? 
```suggestion
               return {
                    ...res,
                    params: {
                        ...params,
                    },
                };
```
To avoid a lot of `else`

Marginy605: it's not convenient for debug (it id often case here) it's more convenient to have one return for that.
I'll think about optimisation

mournfulCoroner (approved):

mournfulCoroner (approved):

PR #415, opened 2023-12-29
Fix error dialog long text

flops:


src/ui/components/DialogErrorWithTabs/DialogErrorWithTabs.scss
```diff
@@ -1,20 +1,33 @@
...
 
+    width: 100%;
+    max-width: $contencDesktopMaxWidth;
+
     &__content {
-        width: 630px;
-        @media (min-width: 320px) and (max-width: 480px) {
-            width: 300px;
+        width: $contencDesktopMaxWidth;
+        @media (max-width: 480px) {
```
Marginy605, line 14: Would it effect on mobile view of dialog content?

cc @mournfulCoroner

src/ui/components/DialogErrorWithTabs/DialogErrorWithTabs.scss
```diff
@@ -1,20 +1,33 @@
...
+            width: $contentSmallScreenWidth;
         }
         max-height: 80vh;
         height: 500px;
         display: flex;
         flex-direction: column;
     }
 
+    &__header {
```
Marginy605, line 23: Why do you use overflow hidden with ellipsis for header? 
Firstly, we have rule that if we use ellipsis than we must add full text as a title to the same element.
Secondly, why can't we just display full caption text? Like this: 
image

Marginy605 (changes requested):

Marginy605:
Let's rename this PR caption. See common style in other's commits https://github.com/datalens-tech/datalens-ui/commits/main/

src/ui/components/DialogErrorWithTabs/DialogErrorWithTabs.scss
flops: I would, currently if screen size is lesser than 320px, modal will jump back to 630px, i suggested that it would be better to remove 'min-width' breakpoint for this case cause even smaller(rear case for 2024) < 320px screen resolution modal will be bigger than on 320px to 480px

mournfulCoroner (approved):

PR #418, opened 2024-01-05
Fix QL chart params default value check, when it is empty string

Flunt1k:
Seems that empty string value is valid, chart returns `No data` in that case, but previously this case was not handled because of simple check for _falsy_ value
```javascript
if (paramDescription && paramDescription.defaultValue) {
    ...
}
```

Since by type defaultValue might be only `string | string[] | {from, to} | undefined` I suggest to check that defaultValue not equal undefined value.


Flunt1k:
@artemipanchuk @kuzmadom please review it

kuzmadom (approved):

artemipanchuk (approved):

PR #425, opened 2024-01-09
Fix for weblate action

Marginy605:


mournfulCoroner (approved):

Marginy605:
Skip tests because it's only affect github action for i18n which is not enabled (used for test flow now)

PR #438, opened 2024-01-10
Fix i18n flow

Marginy605:


Marginy605:
There is no need waiting for tests: changes in PR are concerning only for translations new flow check

mournfulCoroner (approved):

PR #441, opened 2024-01-10
Fix i18n flow bugs

Marginy605:


mournfulCoroner (approved):

Marginy605:
There is no need to wait tests for i18n flow testing

PR #444, opened 2024-01-11
Fix i18n flow

Marginy605:


mournfulCoroner (approved):

Marginy605:
There is no need to wait e2e tests for check i18n os flow

PR #447, opened 2024-01-11
Fix naming for extending value selector

Flunt1k:
Fix naming for function that allows to extend value selector component. 

Semantically it seems more correct to use a hook naming starting with `use`. This also allows redux to be used internally (semantically)


Marginy605 (approved):

Marginy605:
`E2E Tests / Publish E2E Report (pull_request)` all tests have passed, problem is only with publish report. We will solve this later.

PR #451, opened 2024-01-12
Fix i18n flow

Marginy605:


mournfulCoroner (dismissed):

Marginy605:
There is no need to wait tests for check i18n test flow

mournfulCoroner (approved):

PR #452, opened 2024-01-12
Refactoring extending actions for dash control dialog

Flunt1k:


Marginy605 (approved):

Marginy605:
`E2E Tests / Publish E2E Report (pull_request)` all tests have passed, problem is only with publish report. We will solve this later.

PR #469, opened 2024-01-16
Change weblate domain

Marginy605:


mournfulCoroner (approved):

Marginy605:
Config of weblate instance for checking i18n github flow has no affection on e2e tests in TC. No reason to wait them

PR #479, opened 2024-01-17
Add fix to color with gradient by measure values by default

artemipanchuk:


src/server/modes/charts/plugins/datalens/preparers/bar-x/__tests__/prepareBarX.test.ts
```diff
@@ -96,4 +98,10 @@ describe('prepareBarX', () => {
         expect(legendTitles).toEqual(['Color-1', 'Color-2']);
         expect(seriesIds).toEqual(legendTitles);
     });
+
+    test('Measure Values field in colors section -> chart should be by default colored with gradient by measure values', () => {
+        const result = prepareBarX(prepareBarXWithMeasureValuesArgs);
+
+        expect(result.graphs[0].data).toEqual(prepareBarXWithMeasureValuesResult.graphs[0].data);
```
kuzmadom, line 105: There is no color or gradient setting in the result graph right now. If it is the colorValue setting that is being checked in this test, then you need to reflect this in the test name

src/server/modes/charts/plugins/datalens/preparers/bar-x/__tests__/mocks/prepareBarX.mock.ts
```diff
@@ -91,6 +93,148 @@ export const prepareBarXArgs = {
     },
 };
 
+const secondDatasetId = 'y55xt5d3p9csw';
```
kuzmadom, line 96: What is the second dataset for? Why couldn' t use the first one?

kuzmadom (approved):

kuzmadom (approved):

PR #491, opened 2024-01-19
Fix i18n flow

Marginy605:


mournfulCoroner (approved):

Marginy605:
There is no need to wait for e2e test to check i18n flow in weblate

PR #514, opened 2024-01-23
Use outer params instead of merge with local params if outer changed

Marginy605:


src/ui/components/Widgets/Chart/ChartWidget.tsx
```diff
@@ -232,6 +232,35 @@ export const ChartWidget = (props: ChartWidgetProps) => {
         return changedParams;
     }, [prevSavedProps?.params, props.params, usedParamsRef, innerParamsRef, isWizardChart]);
 
+    const clearedOuterParams = React.useMemo(() => {
+        let clearedParams;
+        const propsParams = props.params;
+        const prevSavedPropsParams = prevSavedProps?.params;
+        const isEqualParamsWithPrev = prevSavedProps && isEqual(prevSavedPropsParams, propsParams);
+
+        const changedParams = !prevSavedProps || !isEqualParamsWithPrev;
```
mournfulCoroner, line 241: If it's boolean, maybe you could call it "hasParamsChanged" or something like that

mournfulCoroner (approved):

PR #516, opened 2024-01-23
Add shapes support for d3 visualizations

artemipanchuk:


src/server/modes/charts/plugins/datalens/preparers/line/d3.ts
```diff
@@ -76,6 +76,12 @@ export function prepareD3Line(args: PrepareFunctionArgs): ChartKitWidgetData {
                 },
             },
         };
+
+        if (graph.dashStyle) {
+            d3Graph.dashStyle = graph.dashStyle;
+        }
```
kuzmadom, line 82: Why can't dashStyle be set right away? `undefined` value not supported?

src/server/modes/charts/plugins/datalens/preparers/scatter/d3.ts
```diff
@@ -1,3 +1,4 @@
+import {SymbolType} from '@gravity-ui/chartkit/build/constants';
```
kuzmadom, line 1: Some kind of not very nice import path. I would like to see something like `@gravity-ui/chartkit/d3` here

src/ui/constants/visualizations/scatter.ts
```diff
@@ -118,6 +118,6 @@ export const SCATTER_VISUALIZATION: GraphShared['visualization'] = {
 export const SCATTER_D3_VISUALIZATION: GraphShared['visualization'] = {
     ...SCATTER_VISUALIZATION,
     id: WizardVisualizationId.ScatterD3,
-    allowShapes: false,
+    allowShapes: true,
```
kuzmadom, line 121: I think it is possible to delete a line as in LINE_D3_VISUALIZATION? So that you can see how the visualizations differs

kuzmadom (approved):

kuzmadom (approved):

PR #525, opened 2024-01-25
Fix i18n script launch

Marginy605:


flops (approved):

Marginy605:
There is no need to wait for tests - change in PR is concerning commit create

PR #529, opened 2024-01-25
Add support multiple color fields in Monitoringql charts

Flunt1k:
Add support multiple color fields in colors section for Monitoring charts in QL. 

Since node support multiple fields in color section, we need just add support in UI


Flunt1k:
@artemipanchuk please review it

src/shared/modules/visualization.ts
kuzmadom, line 1: If this empty file is not needed, it is better to delete it

src/ui/units/wizard/components/ValuesList/ValuesList.tsx
```diff
@@ -63,6 +63,7 @@ export interface Props {
     onChangeSelectedValue: (selectedValue: string | null, shouldClearPalette?: boolean) => void;
     extra?: ExtraSettings;
     distincts?: Record<string, string[]>;
+    isMultipleFieldsSupported?: boolean;
```
kuzmadom, line 66: Is there really an additional flag needed here? Can't we rely on the number of items in `items`?

src/ui/units/wizard/components/ValuesList/ValuesList.tsx
```diff
@@ -202,22 +203,26 @@ class ValuesList extends React.Component<Props, State> {
...
-            let distincts;
             let values: string[] = [];
-
             if (externalDistincts) {
-                values = externalDistincts[item.guid];
+                const placeholderFields =
+                    this.props.isMultipleFieldsSupported && items ? items : [item];
+
+                const distincts = placeholderFields.map((v) => {
+                    return externalDistincts[v.guid || v.title];
```
kuzmadom, line 220: I'm not sure if you can rely on the title here. Especially for multiple fields

src/ui/units/wizard/components/ValuesList/ValuesList.tsx
```diff
@@ -388,6 +393,22 @@ class ValuesList extends React.Component<Props, State> {
...
+
+        return root.reduce((acc, v) => {
+            return [
+                ...acc,
+                ...this.buildMultipleColorsDistincts(distincts, index + 1).map(
+                    (v2) => `${v}; ${v2}`,
+                ),
+            ];
+        }, [] as string[]);
+    }
```
kuzmadom, line 411: I think It is necessary to put the algorithm for determining the keys for the dictionary in one place(somewhere in shared) - so that the client and server use the same code

https://github.com/datalens-tech/datalens-ui/blob/e0d63821b25243d9f72fc089c5ffa2265b7bc4e4/src/server/modes/charts/plugins/datalens/preparers/line-time/index.ts#L77

src/ui/units/wizard/components/ValuesList/ValuesList.tsx
Flunt1k: You are right, there is no needs in it. I've added separate prop (`sectionFields`) for this case, which must be used only when multiple fields in section supported.

src/ui/units/wizard/components/ValuesList/ValuesList.tsx
Flunt1k: yes, guid is more than enough in that case

src/ui/units/wizard/components/ValuesList/ValuesList.tsx
Flunt1k: I have extracted the very method of generating legend values. 

Generating all legend elements is a completely different approach on the server and client side. So I believe this should be changed in a different PR and for all preparations.

kuzmadom (approved):

artemipanchuk:
E2E Tests / Publish E2E Report (pull_request) -- tests are ok

PR #548, opened 2024-01-30
Fix save button behaviour for QL charts

Flunt1k:
Now, when a visualization object is loaded in a QL chart, a predefined value is stamped into the repository. However, some fields may not be in the predefined value. 

For this, I added adding fields from the loaded visualization that are not in the predefined visualization

### DIFF
--- a/src/ui/units/ql/utils/visualization/getQlVisualization.ts
+++ b/src/ui/units/ql/utils/visualization/getQlVisualization.ts
@@ -4,9 +4,10 @@ import {Shared} from 'shared';
 import {getAvailableQlVisualizations, getDefaultQlVisualization} from '../visualization';
 
 export const getQlVisualization = (
-    visualizationId: string,
     loadedVisualization: Shared['visualization'],
 ): Shared['visualization'] => {
+    const visualizationId =
+        loadedVisualization.id === 'table' ? 'flatTable' : loadedVisualization.id;
     const availableVisualizations = getAvailableQlVisualizations();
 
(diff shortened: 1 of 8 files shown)


src/ui/units/ql/utils/visualization/getQlVisualization.ts
```diff
@@ -0,0 +1,41 @@
+import mergeWith from 'lodash/mergeWith';
+import {Shared} from 'shared';
+
+import {getAvailableQlVisualizations, getDefaultQlVisualization} from '../visualization';
+
+export const getQlVisualization = (
+    loadedVisualization: Shared['visualization'],
+): Shared['visualization'] => {
+    const visualizationId =
+        loadedVisualization.id === 'table' ? 'flatTable' : loadedVisualization.id;
```
artemipanchuk, line 10: Please, use enum for constants if possible

src/ui/units/ql/utils/visualization/getQlVisualization.ts
```diff
@@ -0,0 +1,41 @@
...
+            candidate,
+            loadedVisualization,
+            (
+                value: any,
+                _sourceValue: any,
+                key,
+                obj: Shared['visualization'] | {},
+                source: Shared['visualization'],
+            ) => {
+                // it needs to prevent override properties which already defined in candidate visualization
```
artemipanchuk, line 27: Please, start comments with capital leter
`// It needs to prevent override properties which already defined in candidate visualization`

artemipanchuk (dismissed):

artemipanchuk:
Tests are ok

PR #573, opened 2024-02-05
Fix i18n flow (instruction md)

Marginy605:


mournfulCoroner (approved):

flops (approved):

Marginy605:
There is no need to wait for tests: only i18n flow was changed

PR #582, opened 2024-02-06
docs: fix tests README port to be consistent with copmpose file

ovsds:


tests/README.md
```diff
@@ -9,7 +9,7 @@
 ### Tests run configuration
 Create `.env` file in root of project. Should be `datalens-ui/.env` and add:
 ```dotenv
-E2E_DOMAIN=http://localhost:3030
```
melikhov-dev, line 12: I think it's more complicated than just changing the port. We will improve this process in the future, but the README is now suitable for front-end development.

PR #588, opened 2024-02-06
Refactoring dash actions and reduces js -> ts p2 

flops:


src/ui/units/dash/containers/Dialogs/Settings/Settings.tsx
```diff
@@ -121,7 +122,10 @@ const Settings = () => {
             dispatch(
                 setSettings({
                     ...settings,
-                    autoupdateInterval: autoupdateInterval || null,
+                    autoupdateInterval:
+                        (typeof autoupdateInterval === 'string'
+                            ? parseInt(autoupdateInterval)
+                            : autoupdateInterval) || null,
```
Marginy605, line 128: Do we really need fallback to null?
I don't like that if we have false in settings we will get null instead

Marginy605 (changes requested):

Marginy605 (approved):

PR #601, opened 2024-02-12
Split editor type config tabs to separate constants

Marginy605:


Marginy605:
image
Locally failed test passed 
image
flacky problem

PR #631, opened 2024-02-16
Correct handle of no data case in markup wizard chart

artemipanchuk:


artemipanchuk:
Cannot create valid case for unit test, this issue originally was detected on client side with ChartKit

artemipanchuk:
Related to https://github.com/datalens-tech/datalens/issues/92

PR #667, opened 2024-02-22
Fix publish platform external contrib

goshander:


Marginy605:
No need to wait for e2e tests, only publish platform ci was changed in PR

PR #686, opened 2024-02-27
Fix old ql metric charts render with markup

artemipanchuk:


src/server/modes/charts/plugins/ql/preparers/metric.ts
```diff
@@ -98,32 +108,61 @@ export default ({
...
+                        {
+                            type: 'color',
+                            color,
+                            content: {
+                                className: 'markup-metric-value',
+                                type: 'text',
+                                content: formattedValue,
+                            },
+                        },
+                    ],
```
kuzmadom, line 145: It looks almost identical to https://github.com/datalens-tech/datalens-ui/blob/31782815a7be290a5760be2a765b61768bf0bfab/src/server/modes/charts/plugins/datalens/preparers/metric/variants/markup.ts#L88
It would be healthy to share the code somehow

kuzmadom (approved):

artemipanchuk:
Related to https://github.com/datalens-tech/datalens/issues/92

PR #696, opened 2024-02-29
Rework DialogAddYadoc component

korvin89:
- rework ya document adding dialog
- add `updateDialogProps` method to `DialogManager`


src/ui/store/reducers/dialog.ts
```diff
@@ -23,6 +23,32 @@ function dialog(state: DialogState = getInitialDialogState(), action: DialogActi
             };
         }
 
+        case UPDATE_DIALOG_PROPS: {
+            const {id, props} = action;
+            const dialogToUpdateIndex = state.dialogs.findIndex((d) => d.id === id);
```
Marginy605, line 28: One letter is not a good naming practice, let's rename to something meaninfull

Marginy605: Discussed this case. Let it d.

Marginy605 (approved):

PR #733, opened 2024-03-07
Update statoscope.config.js

melikhov-dev:
disable lodash error, enable moment


melikhov-dev:
don't need run E2E tests

PR #775, opened 2024-03-14
Add table component based on @tanstack/react-table

kuzmadom:
Related to https://github.com/datalens-tech/datalens/issues/88

### DIFF
--- a/src/ui/components/Table/Table.tsx
+++ b/src/ui/components/Table/Table.tsx
@@ -1,10 +1,7 @@
 import React from 'react';
 
+import type {SortingState} from '@tanstack/react-table';
 import {
-    ColumnDef,
-    SortingState,
-    createColumnHelper,
-    flexRender,
     getCoreRowModel,
     getGroupedRowModel,
     getSortedRowModel,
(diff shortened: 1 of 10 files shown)


src/ui/components/Table/Table.tsx
```diff
@@ -0,0 +1,321 @@
...
+        }
+    };
+
+    const tableRows = table.getRowModel().rows;
+    const tableHeaders = table.getHeaderGroups();
+    const shouldShowFooter = columns.some((column) => column.footer);
+
+    return (
+        <div className={b()}>
+            {title && <div className={b('title')}>{title.text}</div>}
```
korvin89, line 182: Maybe we should use [native caption](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption)? Then we will need no more a `div` wrapper

src/ui/components/Table/Table.tsx
```diff
@@ -0,0 +1,321 @@
...
+            <table className={b('table')}>
+                <thead
```
korvin89, line 184: I suggest you to separate `thead`, `tbody` and `tfoot` renders

src/ui/components/Table/types.ts
```diff
@@ -0,0 +1,68 @@
...
+        footer?: TFoot[];
+    };
+    noData?: {
+        text: string;
+    };
+    header?: {
+        sticky?: boolean;
+    };
+    onClick?: OnTableClick;
+    qa?: {
```
korvin89, line 59: It looks complicated, besides all of such properties in gravity are strings and it will be inconsistent. I think it is quite enough to have one `qa` attribute on table itself

korvin89 (approved):

PR #781, opened 2024-03-15
Fix columns sorting for rows with markup

kuzmadom:


src/shared/modules/markup.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import {MarkupItem} from '../types';
+
+export function isMarkupItem(obj: unknown): obj is MarkupItem {
+    return isPlainObject(obj) && isMarkupObject(obj as object);
+}
+
+function isMarkupObject(obj: {type?: string}): boolean {
+    if (
+        (obj.type === 'concat' ||
```
korvin89, line 11: Should we add all of [these](https://github.com/datalens-tech/datalens-ui/blob/main/src/ui/components/Markup/constants.ts#L1) types?

korvin89 (approved):

PR #811, opened 2024-03-25
Add resetting shapes palette after visualization type change

artemipanchuk:


src/ui/units/wizard/components/Dialogs/DialogShapes/DialogShapes.tsx
```diff
@@ -59,9 +58,13 @@ const DialogShapes: React.FC<Props> = ({
...
 }: Props) => {
     const [shapesState, setShapesState] = React.useState<ShapesState>({
+        paletteType,
         selectedValue: null,
         mountedShapes:
-            shapesConfig.mountedShapes && Object.keys(shapesConfig.mountedShapes)
+            shapesConfig.mountedShapes &&
+            Object.keys(shapesConfig.mountedShapes) &&
+            (typeof shapesConfig.paletteType === 'undefined' ||
+                shapesConfig.paletteType === paletteType)
```
kuzmadom, line 67: This won't fix the error for the current saved charts, right?
It is the values stored in mountedShapes that need to be validated, not just the type of config. Because there may be a situation with the correct config type, but an incorrect value in mountedShapes

src/ui/units/wizard/components/Dialogs/DialogShapes/DialogShapes.tsx
```diff
@@ -96,11 +99,12 @@ const DialogShapes: React.FC<Props> = ({
     }, [onClose]);
 
     const onApplyButtonClick = React.useCallback(() => {
-        const shapesConfig: ShapesConfig = {
+        onApply({
             mountedShapes: shapesState.mountedShapes,
             fieldGuid: item.guid,
-        };
-        onApply(shapesConfig);
+            paletteType: shapesState.paletteType,
```
kuzmadom, line 105: The type of palette can be determined by the type of visualization. This is a calculated field, it does not need to be stored in the config

PR #818, opened 2024-03-26
Added telemetry for js execution timeouts

melikhov-dev:


src/server/components/charts-engine/components/processor/index.ts
```diff
@@ -1084,6 +1085,12 @@ export class Processor {
                         statusCode: DEFAULT_RUNTIME_TIMEOUT_STATUS,
                     };
 
+                    onCodeExecuted({
```
kuzmadom, line 1088: This will be run on timeout on any tab, right? Probably only needed for js?

melikhov-dev: Yes, but this is normal, because we are interested in all cases of timeouts, and the separation is difficult. But this can be changed after refactoring.

kuzmadom (approved):

PR #821, opened 2024-03-27
fix % symbol in key

melikhov-dev:
`url.parse()` does not replace `%` with `%25`, but `encodeURIComponent` does. What's why we need to replace all 'real' `%` with `%25` manually


src/server/modes/charts/plugins/runner-key-adapter.ts
```diff
@@ -32,7 +32,9 @@ export const plugin = {
 
                             // Otherwise, the request with a space will be encoded, and the request in US will encode the encoded space
                             // Datalens%20Charts/Errors/no-data -> Datalens%2520Charts%2FErrors%2Fno-data
-                            const decodedPathname = decodeURIComponent(parsedKey.pathname!);
+                            const decodedPathname = decodeURIComponent(
+                                parsedKey.pathname!.replace(/%(?![0-9][0-9a-fA-F]+)/g, '%25'),
```
jhoncool, line 36: It doesn't work for case `some%1some`

jhoncool (approved):

artemipanchuk (approved):

PR #904, opened 2024-04-17
Fix term popup bg

Marginy605:


flops (approved):

Marginy605:
There is no need to wait for e2e test for color var changing

PR #910, opened 2024-04-18
Fix Inspector dialog styles

korvin89:


src/ui/libs/DatalensChartkit/components/ChartKitBase/components/Header/components/Menu/Items/Inspector/Inspector.scss
```diff
@@ -14,26 +15,12 @@
         margin-bottom: 20px;
     }
 
-    &__request-id {
-        color: var(--g-color-text-secondary);
```
korvin89, line 18: This element should have accent color case it is form label

imsitnikov (approved):

PR #938, opened 2024-04-25
Hide duplicate btn for workbook entries in flat lists

Sergey-weber:


src/ui/components/EntryContextMenu/constants.ts
```diff
@@ -38,7 +38,11 @@ const CONTEXT_MENU_COPY = {
             : undefined;
     },
     // remain Copy menu item in navigation, but show in actionPanel only if actual version is opened
-    isVisible({showSpecificItems}: ContextMenuParams) {
+    isVisible({showSpecificItems, entry}: ContextMenuParams) {
+        if (entry?.workbookId) {
```
imsitnikov, line 42: It also hides "Duplicate" button on workbook entity page.

The action should remain on workbook entity page:
Screenshot 2024-04-25 at 12 46 14

imsitnikov (changes requested):

src/ui/components/EntryContextMenu/constants.ts
Sergey-weber: added flag - isFlatList for hide only in flat list place

imsitnikov (approved):

PR #964, opened 2024-05-06
Support reload event and render notification for embedded charts

mournfulCoroner:


src/ui/components/Widgets/Chart/Chart.tsx
```diff
@@ -216,6 +219,26 @@ export const Chart = (props: ChartNoWidgetProps) => {
...
+    React.useEffect(() => {
+        const isEmbedded = isEmbeddedMode();
+        const handleMessageEvent = (event: MessageEvent) => {
+            if (event.data.type === CHART_RELOAD_EVENT) {
+                reloadChart();
+            }
+        };
+
+        if (isEmbedded) {
+            window.addEventListener('message', handleMessageEvent);
```
flops, line 232: Maybe move return here and do not check on every reloadChart ref update it for no reason?

flops (approved):

PR #975, opened 2024-05-08
Fix `MetricWidget` rendering after data updating

korvin89:


src/ui/libs/DatalensChartkit/ChartKit/plugins/Metric/renderer/MetricTile.tsx
```diff
@@ -115,9 +115,9 @@ export class MetricTile extends React.PureComponent<MetricTileProps> {
         }
 
         const {config, callback} = getGraph(chartConfig, chart, undefined, isMobile);
-
         return (
             <HighchartsReact
+                key={Math.random()}
```
korvin89, line 120: We already use such hack [here](https://github.com/gravity-ui/chartkit/blob/main/src/plugins/highcharts/renderer/components/HighchartsComponent.tsx#L162)

jhoncool (approved):

PR #976, opened 2024-05-08
Add e2e-tests for group controls

mournfulCoroner:
New tests: 
1. 3 tests to check the work of the "Apply" and "Reset" buttons.
2. Checking the transformation of the selector from the old single format to the new one, then adding and removing a selector from group.
3. Checking for changing the order of selectors from placement settings.


tests/opensource-suites/dash/selectors/groupBaseActions.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    SECOND_CONTROL: {
+        controlTitle: 'test-control-2',
+        controlItems: ['1', '2'],
+        controlFieldName: 'test-control-field-2',
+    },
+    OLD_CONTROL_TITLE: 'city',
+    OLD_SELECTOR_TAB: 'Tab 2',
+};
+
+const getSelectorsLabels = async (page: Page) => {
```
Marginy605, line 32: May be it's better to add this method in DashboardPage class?

Marginy605 (changes requested):

tests/opensource-suites/dash/selectors/groupButtonActions.test.ts
```diff
@@ -0,0 +1,261 @@
...
+        controlTitle: 'test-control-2',
+        controlItems: ['1', '2'],
+        controlFieldName: 'test-control-field-2',
+        defaultValue: '2',
+    },
+};
+
+const STATE_CHANGING_TIMEOUT = 1000;
+
+const getNewStateHash = async (page: Page, action: () => Promise<void>) => {
```
Marginy605, line 27: The same: perhaps it's better to move to common class

tests/opensource-suites/dash/selectors/groupButtonActions.test.ts
```diff
@@ -0,0 +1,261 @@
...
+        async ({page}: {page: Page}) => {
+            const dashboardPage = new DashboardPage({page});
+
+            await page.locator(slct(DialogGroupControlQa.applyButtonCheckbox)).click();
+
+            await page.locator(slct(ControlQA.dialogControlApplyBtn)).click();
+            await dashboardPage.saveChanges();
+
+            // check that there's no state
+            expect(getUrlStateParam(page)).toEqual(null);
```
Marginy605, line 87: It's not a good idea to check for absence. If there would be state later this condition would be passed although.

tests/opensource-suites/dash/selectors/groupButtonActions.test.ts
```diff
@@ -0,0 +1,261 @@
...
+                {title: PARAMS.FIRST_CONTROL.controlTitle},
+                PARAMS.FIRST_CONTROL.controlItems[0],
+            );
+
+            // timeout so state has time to be wrote in search params
+            await page.waitForTimeout(STATE_CHANGING_TIMEOUT);
+            // check that there's no state (apply button isn't pressed after changing value)
+            expect(getUrlStateParam(page)).toEqual(null);
+
+            const requestPromise = page.waitForRequest('/gateway/root/us/createDashState');
```
Marginy605, line 109: It's better to use constant for such url. As far as I remember, we have already constant for that (for api/run), but we don't use it (in no reason)

src/ui/libs/DatalensChartkit/components/Control/Items/Items.js
```diff
@@ -488,10 +489,7 @@ BaseControlButton.propTypes = {
     label: PropTypes.string,
     theme: PropTypes.string,
     onChange: PropTypes.func.isRequired,
-    className: PropTypes.string,
```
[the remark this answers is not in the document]
mournfulCoroner: Even if you pass `className` in `BaseControlButton` it will be lost [there](https://github.com/datalens-tech/datalens-ui/pull/976/files/09ead15dc784a63c0acae7d1edc94a20f4d9d4c8#diff-fa9b3120811aeee9cca6f9c03bae279959a8c44acb511c19371b7a9588fe5aa9L465) and will not be passed to `Button`

flops (approved):

src/ui/libs/DatalensChartkit/components/Control/Items/Items.js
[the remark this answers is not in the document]
mournfulCoroner: We decided to leave the deletion

Marginy605 (approved):

PR #1049, opened 2024-05-28
Improve performance for flat tables

melikhov-dev:


src/server/modes/charts/plugins/datalens/preparers/flat-table/index.ts
```diff
@@ -290,7 +309,7 @@ function prepareFlatTable({
             }
 
             if (isActionParamsEnable) {
-                if (canUseFieldForFiltering(item)) {
+                if (item.canUseFieldForFiltering) {
                     if (isDateField(item)) {
```
melikhov-dev, line 313: @kuzmadom is ok to using `isDateField` in this place?  Why not `isDateType`?

kuzmadom: there is not much difference what to use here

`isDateField` inside uses `isDateType`

kuzmadom (dismissed):

PR #1065, opened 2024-05-31
Fix ignore nulls behaviour

artemipanchuk:


src/server/modes/charts/plugins/datalens/preparers/line/prepare-line-data.ts
```diff
@@ -346,6 +347,10 @@ export function prepareLineData(args: PrepareFunctionArgs) {
                                 value = 0;
                             }
 
+                            if (typeof value === 'undefined' && nulls === 'ignore') {
+                                value = null;
+                            }
```
kuzmadom, line 352: below, `value` already transform to null
Can you move the check below to `!isXCategoryAxis && typeof value === 'undefined' && nulls === 'connect'`? To avoid doing unnecessary work

kuzmadom (approved):

PR #1106, opened 2024-06-10
Fix cancellation of request in charts and selectors

mournfulCoroner:
- In single and group selectors, requests are canceled if another request is sent.
- Requests for charts and selectors are canceled on unmount.

### DIFF
--- a/src/ui/components/DashKit/plugins/GroupControl/Control/Control.tsx
+++ b/src/ui/components/DashKit/plugins/GroupControl/Control/Control.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 
 import {Loader} from '@gravity-ui/uikit';
+import type {CancelTokenSource} from 'axios';
+import axios from 'axios';
 import block from 'bem-cn-lite';
 import {I18n} from 'i18n';
 import isEqual from 'lodash/isEqual';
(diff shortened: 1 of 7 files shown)


src/ui/components/DashKit/plugins/GroupControl/Control/Control.tsx
```diff
@@ -98,10 +99,10 @@ export const Control = ({
     getDistincts,
     onChange,
     needReload,
-    cancelSource,
     workbookId,
 }: ControlProps) => {
     const [prevNeedReload, setPrevNeedReload] = React.useState(needReload);
+    const [requestCancellation, setRequestCancellation] = React.useState<CancelTokenSource>();
```
flops, line 105: Do we really need it as state? I don't see any needs for component to redraw while this object updates, why not use ref?

src/ui/components/Widgets/Chart/hooks/useLoadingChart.ts
```diff
@@ -534,6 +534,20 @@ export const useLoadingChart = (props: LoadingChartHookProps) => {
         enable: !isInit && !canBeLoaded,
     });
 
+    /**
+     * cancel active requests when component is unmounted (changing dash tab, switching to another page)
+     */
+    React.useEffect(() => {
```
flops, line 540: Why we duplicating this logic ?
https://github.com/datalens-tech/datalens-ui/blob/f7953bb11cbc6227105f3c00bcd832ea43add40d/src/ui/components/Widgets/Chart/hooks/useLoadingChart.ts#L319-L320

flops (approved):

PR #1114, opened 2024-06-11
Add hotkeys for undo/redo in wizard

artemipanchuk:


src/ui/units/wizard/containers/App.tsx
```diff
@@ -24,9 +25,15 @@ interface Props extends StateProps, DispatchProps {}
 const App = ({widget, setCurrentPageEntry, asideHeaderData, ...routeProps}: Props) => {
     const isAsideHeaderEnabled = getIsAsideHeaderEnabled();
 
+    const hotkeysContext = useHotkeysContext();
+
     React.useEffect(() => {
+        hotkeysContext.enableScope('wizard');
```
flops, line 31: Let's make 'wizard' - as constants

src/ui/units/wizard/containers/App.tsx
```diff
@@ -24,9 +25,15 @@ interface Props extends StateProps, DispatchProps {}
 const App = ({widget, setCurrentPageEntry, asideHeaderData, ...routeProps}: Props) => {
     const isAsideHeaderEnabled = getIsAsideHeaderEnabled();
 
+    const hotkeysContext = useHotkeysContext();
```
flops, line 28: It would be great to make it as external enable\disable hook for re-use

src/ui/units/wizard/containers/Wizard/WizardActionPanel/useWizardActionPanel.ts
```diff
@@ -47,6 +48,28 @@ export const useWizardActionPanel = (
...
+    };
+
+    const onClickGoForward = () => {
+        dispatch(goForward({unitId: WIZARD_EDIT_HISTORY_UNIT_ID}));
+    };
+
+    let undoHotkey;
+    let redoHotkey;
+
+    if (isMacintosh()) {
```
flops, line 62: Would be great to have it as hook too.
Best what i could think of is shared part for meta\ctrl detect and declarative (object like) config for init:

```
useHotkeys([
  {
    key: [SOME_CONST_META_KEY, 'shift', 'z'],
    handler: onClickGoBack,
    options: { scopes: WIZARD_SCOPE_CONST },
  }
])
```

src/ui/components/DialogManager/DialogManagerContainer.tsx
```diff
@@ -1,17 +1,27 @@
...
+import {closeDialog} from '../../store/actions/dialog';
+import history from '../../utils/history';
 
 import DialogManager from './DialogManager';
 
 export const DialogManagerContainer = () => {
     const dispatch = useDispatch();
     const dialogs = useSelector((state: DatalensGlobalState) => state.dialog.dialogs);
 
+    const hotkeysContext = useHotkeysContext();
```
flops, line 18: Do we plan to change hotkeyContext? Why not move `const hotkeysContext = useHotkeysContext();` inside useSetHotkeysScope ?

src/ui/units/wizard/containers/Wizard/WizardActionPanel/useWizardActionPanel.ts
```diff
@@ -47,6 +49,26 @@ export const useWizardActionPanel = (
 
     const enableEditHistory = Utils.isEnabledFeature(Feature.EnableEditHistory);
 
+    const onClickGoBack = () => {
```
flops, line 52: `onClickGoBack` and `onClickGoForward` are internally created functions, if we pass this as dependencies to `useMemo` - it will call it every time render is called. So should we use useCallback or save it to ref ourself or don't use useMemo with it.

flops (approved):

PR #1157, opened 2024-06-21
Fix crash group controls after adding select

mournfulCoroner:


mournfulCoroner:

mournfulCoroner:
A flapping test has failed. The code in the pr could not affect him

PR #1163, opened 2024-06-24
Add warning about unused items

artemipanchuk:
Potentially dangerous case. Warning is visible:
image

Safe case. No warning.
image


src/server/modes/charts/plugins/datalens/preparers/line/helpers/line-mappers/types.ts
```diff
@@ -154,4 +154,8 @@ export interface ExtendLineWithSegmentDataArgs {
 
 export type ColorAndShapesValues = {colorValue?: string; shapeValue?: string};
 
-export type MappedLinesKeys = {key: string; lastKey: string | number | Date | null};
+export type MappedLinesKeys = {
+    key: string;
+    lastKey: string | number | Date | null;
+    dataConflict?: boolean;
```
kuzmadom, line 160: Сan you rename it to something like "uniqXValues: false" or "hasDublicatesByX: true"? `dataConflict` is a very vague name

src/server/modes/charts/plugins/ql/js.ts
```diff
@@ -313,7 +313,10 @@ export default ({shared, ChartEditor}: {shared: QlConfig; ChartEditor: IChartEdi
                 distincts: resultDistincts,
             };
         } else {
+            const dataConflict = result.graphs.some((graph: any) => graph.dataConflict);
```
kuzmadom, line 316: You need to add a check for the existence of graphs - in some cases, there may be no data on the chart

kuzmadom (approved):

PR #1172, opened 2024-06-25
Enable menu option showing related objects of current object

mournfulCoroner:


mournfulCoroner:
Related to https://github.com/datalens-tech/datalens/issues/81

Marginy605 (approved):

mournfulCoroner:
The failure of the tests is not related to the code in the pr, but to the fact that one of the connections from these tests temporarily does not work

PR #1185, opened 2024-06-26
Support changing title on double click in group controls & up uikit@6.20.0

mournfulCoroner:


mournfulCoroner:

PR #1190, opened 2024-06-27
Fix case with apply and aliases in group control

mournfulCoroner:


mournfulCoroner:

PR #1194, opened 2024-06-28
Add configurable functions with navigation settings in registry

artemipanchuk:


src/ui/components/Navigation/Base/configure.tsx
```diff
@@ -0,0 +1,293 @@
...
+                        ? undefined
+                        : () => {
+                              onClick(targetPlace.value);
+                          }
+                }
+            >
+                {targetPlace.buttonText}
+            </Button>
+        );
+    },
```
kuzmadom, line 292: It would be good to split this file into two parts - react components and utilities.
And it's a bit strange to see the components in the `configure` file

kuzmadom (approved):

PR #1199, opened 2024-06-29
Fix for 401 on parallel invalid access token renewal

stankis:


src/ui/libs/schematic-sdk/index.ts
```diff
@@ -39,6 +40,14 @@ export type DatalensSdk<TSchema extends SchemasByScope> = (<T>(
 
 const sdkConfig: SdkConfig = {
     csrfToken: Utils.getCSRFToken() || '',
+    axiosConfig: {
```
resure, line 43: What would delay be equal to for this retries?

stankis: discussed offline, added

src/ui/libs/DatalensChartkit/modules/axios/axios.ts
```diff
@@ -2,6 +2,7 @@ import type {AxiosError} from 'axios';
 import axios from 'axios';
 import axiosRetry, {isRetryableError} from 'axios-retry';
 import isNumber from 'lodash/isNumber';
+import {sleep} from 'shared';
```
korvin89, line 5: Lets grab this helper like `shared/modules` to avoid global indexe problem

src/ui/libs/DatalensChartkit/modules/axios/axios.ts
```diff
@@ -42,7 +43,9 @@ export function initConcurrencyManager(maxConcurrentRequests: number) {
 
     axiosRetry(client, {
         retries: 0,
-        retryCondition: isRetryableError,
+        retryCondition: (error) => {
```
korvin89, line 46: Old version is suitable as well

src/ui/libs/DatalensChartkit/modules/axios/axios.ts
```diff
@@ -55,11 +58,22 @@ export function initConcurrencyManager(maxConcurrentRequests: number) {
 
 client.interceptors.response.use(
     (data) => data,
-    (error) => {
+    async (error) => {
         if (error?.response?.status === 451) {
             showReadOnlyToast();
         }
 
+        if (error && error.response && error.response.status === 498) {
```
korvin89, line 66: I suggest you to use `axios.isAxiosError(error)` here

src/ui/libs/DatalensChartkit/modules/axios/axios.ts
stankis: agree, fixed

src/ui/libs/DatalensChartkit/modules/axios/axios.ts
stankis: sure :)

src/ui/libs/DatalensChartkit/modules/axios/axios.ts
stankis: agree, added

korvin89 (approved):

PR #1244, opened 2024-07-11
Add tooltip support in generateHtml

korvin89:


src/ui/libs/DatalensChartkit/ChartKit/components/ChartKitTooltip/ChartKitTooltip.tsx
```diff
@@ -0,0 +1,135 @@
...
+    return id || '';
+};
+
+const getTooltipContent = (value = '') => {
+    let result = value;
+
+    try {
+        const json = JSON.parse(value);
+        result = generateHtml(json);
+    } catch {}
```
kuzmadom, line 31: I think there is no need to swallow an error here - it will be difficult to figure out why the json does not turn into correct markup when using invalid tags, for example

src/ui/libs/DatalensChartkit/ChartKit/components/ChartKitTooltip/ChartKitTooltip.tsx
```diff
@@ -0,0 +1,135 @@
...
+    if (!Object.keys(anchors).length) {
+        return null;
+    }
+
+    return (
+        <React.Fragment>
+            {Object.entries(anchors).map(([id, anchor]) => {
+                return (
+                    <Popup key={id} anchorRef={anchor.ref} open={anchor.open}>
+                        <div dangerouslySetInnerHTML={{__html: anchor.content}} />
```
kuzmadom, line 127: Maybe it's worth adding styles for the container? I would like to limit the maximum width and add padding so that a сonsistent style with the service is maintained.

src/ui/libs/DatalensChartkit/ChartKit/components/ChartKitTooltip/ChartKitTooltip.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    if (!Object.keys(anchors).length) {
+        return null;
+    }
+
+    return (
+        <React.Fragment>
+            {Object.entries(anchors).map(([id, anchor]) => {
+                return (
+                    <Popup key={id} anchorRef={anchor.ref} open={anchor.open}>
```
kuzmadom, line 126: Can you add other properties of the tooltip? For example, the `placement` and `hasArrow`

src/ui/libs/DatalensChartkit/modules/html-generator/index.ts
```diff
@@ -25,7 +32,7 @@ export function generateHtml(item?: ChartKitHtmlItem | ChartKitHtmlItem[] | stri
             });
         }
 
-        const elem = document.createElement(tag);
+        const elem = document.createElement(tag === TAG_DL_TOOLTIP ? 'div' : tag);
```
kuzmadom, line 35: Is it possible to add the ability to specify a different tag?
When using div, additional styles will be needed when used in the text

kuzmadom (approved):

kuzmadom (approved):

PR #1250, opened 2024-07-11
Add sandbox engine selector

melikhov-dev:
Just a temporary selector for the experimental period


src/ui/units/editor/store/reducers/editor/editor.js
```diff
@@ -384,6 +385,10 @@ export const getScriptsValues = (state) => state.editor.scriptsValues;
 export const getIsScriptsChanged = createSelector(
     [getEntry, getScriptsValues],
     (entry, scriptsValues) => {
+        if (entry.meta.is_sandbox_version_changed) {
```
kuzmadom, line 388: `entry.meta?.is_sandbox_version_changed`

`meta` may not be defined

melikhov-dev: Yep, fixed it

PR #1251, opened 2024-07-11
Add rows render limit for charts

flops:


src/ui/components/Widgets/Chart/ChartWidget.tsx
```diff
@@ -485,6 +486,13 @@ export const ChartWidget = (props: ChartWidgetProps) => {
         [context.canEdit, editMode, context.entryDialoguesRef, initName],
     );
 
+    const dashkitConfig = React.useMemo(
```
kuzmadom, line 489: Judging by what lies in the object, I would expect something like dashState instead of dashkitConfig (which is like a common dashboard setting regardless of the state)

src/ui/components/Widgets/Chart/Chart.tsx
```diff
@@ -281,6 +282,7 @@ export const Chart = (props: ChartNoWidgetProps) => {
                 isWidgetMenuDataChanged={isWidgetMenuDataChanged}
                 renderPluginLoader={renderPluginLoader}
                 paneSplitOrientation={paneSplitOrientation}
+                dashkitConfig={dashkitConfig}
```
kuzmadom, line 285: Now the setting initiates the rerender of all charts - regardless of the need. Not very good

kuzmadom (approved):

PR #1254, opened 2024-07-12
Fix chartkit tooltip in case of markup in anchor

korvin89:
There was no trigger event in case of using markap as a content of tooltips's anchor


src/ui/libs/DatalensChartkit/ChartKit/components/ChartKitTooltip/ChartKitTooltip.tsx
```diff
@@ -53,12 +53,18 @@ const ChartKitTooltipComponent = React.forwardRef<ChartKitTooltipRef | undefined
...
+                    let node = e.target as HTMLElement | null;
 
                     if (!node) {
                         return;
                     }
 
+                    const anchorId = node.dataset['tooltipAnchorId'];
+
+                    if (anchorId) {
+                        node = document.getElementById(anchorId) || node;
```
kuzmadom, line 65: Perhaps it was worth looking for a parent element with a specific attribute(data-tooltip-content)? Then you wouldn't have to add additional attributes to all child nodes

kuzmadom (approved):

PR #1317, opened 2024-08-05
Adds a percentage enabling of a new sandbox

melikhov-dev:
We need to have a mechanism for a percentage enabling  new sandbox. By default it needs to be false.

### DIFF
--- a/src/server/components/charts-engine/runners/editor.ts
+++ b/src/server/components/charts-engine/runners/editor.ts
@@ -14,6 +14,27 @@ import {prepareErrorForLogger} from './utils';
 
 import type {RunnerHandlerProps} from '.';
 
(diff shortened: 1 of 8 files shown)


src/server/components/charts-engine/runners/editor.ts
```diff
@@ -14,6 +14,27 @@ import {prepareErrorForLogger} from './utils';
 
 import type {RunnerHandlerProps} from '.';
 
+const NEW_SANBOX_PERCENT = {
```
imsitnikov, line 17: SAN**D**BOX

imsitnikov (approved):

jhoncool (dismissed):

jhoncool (approved):

imsitnikov (approved):

PR #1373, opened 2024-08-15
Additional libs in the sandbox

kuzmadom:


melikhov-dev:
I don't like this 
src/ui/libs/DatalensChartkit/modules/ui-sandbox/libs/date-utils.v2.4.0.js

because I have no idea what's inside this package and where I can find the sources, fix and reproduce the build.

melikhov-dev (approved):

PR #1388, opened 2024-08-21
Use /v1/structure-items endpoint

TheXardas:
- Allows for single list pagination, instead of collections and workbooks juggling
- Redux boilerplate is reused from old endpoint - /v1/collection-content
- Old endpoint should be removed after successful integration


src/shared/schema/us/actions/collections.ts
```diff
@@ -59,8 +59,7 @@ export const collectionsActions = {
                 collectionId,
                 includePermissionsInfo,
                 filterString,
```
imsitnikov, line 61: Let's rename the action to `getStructureItems`, renaming is not necessary for actions in Redux, but here it is better to do because it will be obvious in Network tab what method is called in us.

imsitnikov (changes requested):

jhoncool (approved):

PR #1406, opened 2024-08-24
Add dash control runner

Marginy605:


Marginy605:
Linked to https://github.com/datalens-tech/ui-yandex/pull/1959

src/server/components/charts-engine/components/processor/index.ts
```diff
@@ -663,14 +669,23 @@ export class Processor {
...
+                } else {
+                    jsTabResults = await builder.buildChart({
+                        data,
+                        sources: resolvedSources,
+                        params,
+                        usedParams,
+                        actionParams: normalizedActionParamsOverride,
+                        hooks,
+                    });
+                }
```
kuzmadom, line 687: Why not make some of the arguments optional? Then you won't need a fork for different types

src/server/components/charts-engine/components/processor/control-api-context.ts
```diff
@@ -0,0 +1,102 @@
...
+        updateParams: () => {},
+        updateActionParams: () => {},
+        updateLibraryConfig: () => {},
+        setSideHtml: () => {},
+        setSideMarkdown: () => {},
+        setExportFilename: () => {},
+        wrapFn: () => ({}) as ReturnType<IChartEditor['wrapFn']>,
+        generateHtml: () => ({}) as ReturnType<IChartEditor['generateHtml']>,
+        attachHandler: () => ({}) as ReturnType<IChartEditor['attachHandler']>,
+        attachFormatter: () => ({}) as ReturnType<IChartEditor['attachFormatter']>,
```
kuzmadom, line 55: If a common `IChartEditor` interface is used anyway, then I would like to use a common method - now there is an extra duplication that you can stumble over

Marginy605: I've removed it and use chart-api-context

kuzmadom (dismissed):

src/server/components/charts-engine/components/processor/control-builder.ts
```diff
@@ -0,0 +1,150 @@
...
+        buildChartLibraryConfig: emptyStep('Highcharts'),
+
+        buildChartConfig: emptyStep('Config'),
+
+        buildChart: async (options) => {
+            const {data, params} = options;
+            const timeStart = process.hrtime();
+
+            const context = getChartApiContext({
+                name: 'JavaScript',
```
melikhov-dev, line 103: It will be nice to create const enum for tab names. I will do it by myself later

src/server/components/charts-engine/runners/control.ts
```diff
@@ -0,0 +1,226 @@
...
+
+                            if ('_confStorageConfig' in resultCopy) {
+                                delete resultCopy._confStorageConfig;
+                            }
+
+                            const logError = prepareErrorForLogger(result.error);
+
+                            cx.log('PROCESSED_WITH_ERRORS', {error: logError});
+
+                            let statusCode = 500;
```
melikhov-dev, line 139: This one (and  below) code is the same for all runners. It would be nice to extract it into an external function and reuse it.

Because it's to complicate for support as copy. There are so many places for bugs.

src/server/components/charts-engine/components/processor/control-builder.ts
Marginy605: Discussed to fix it later at whole usages

kuzmadom (approved):

melikhov-dev (approved):

PR #1490, opened 2024-09-09
Editor menu fix

stpavlenko:


src/ui/units/editor/components/Preview/Preview.js
```diff
@@ -26,6 +27,10 @@ const Preview = ({
         onLoadData({data: result.data, status});
     };
 
+    const getMenuType = () => {
```
Marginy605, line 30: Why function? It's better use const with condition

Marginy605 (approved):

PR #1530, opened 2024-09-16
Fix empty ql dialog

stpavlenko:


src/ui/units/wizard/containers/Wizard/SectionVisualization/VisualizationItem/VisualizationItem.tsx
```diff
@@ -333,7 +338,10 @@ class VisualizationItem extends React.Component<Props, State> {
                 <div
                     className={`item-icon ${item.data_type}-icon`}
                     onClick={(event) => {
-                        if (isParameter(item)) {
+                        const isEmptyQaDialog =
+                            this.props.qlChartType && emptyQaDialogTypes.includes(item.data_type);
```
kuzmadom, line 342: You can't rely on that
In some visualizations, date formatting is used, and the table has settings for all fields - for example, a hint

kuzmadom (changes requested):

src/ui/units/wizard/components/Dialogs/DialogField/DialogField.tsx
```diff
@@ -267,6 +271,12 @@ class DialogField extends React.PureComponent<DialogFieldInnerProps, DialogField
                 !Number.isNaN((formatting as CommonNumberFormattingOptions).precision);
         }
 
+        const modalBody = this.renderModalBody();
+        if (!modalBody) {
+            this.props.onCancel();
```
kuzmadom, line 276: Can you add a comment here why do I need to call cancel? So as not to be deleted accidentally later

src/ui/units/wizard/components/Dialogs/DialogField/DialogField.tsx
```diff
@@ -4,6 +4,7 @@ import {Dialog, Icon, Switch} from '@gravity-ui/uikit';
 import block from 'bem-cn-lite';
 import DialogManager from 'components/DialogManager/DialogManager';
 import {i18n} from 'i18n';
+import {isNull} from 'lodash';
```
kuzmadom, line 7: `import isNull from 'lodash/isNull';`

PR #1569, opened 2024-09-19
QL save dialog before screenshot

stpavlenko:


src/ui/units/ql/store/actions/ql.ts
```diff
@@ -1185,6 +1185,14 @@ export const updateChart = (data: QlConfig, mode?: EntryUpdateMode) => {
     };
 };
 
+export const updateChartAndDoAction = (preparedChartData: QlConfig, action: () => void) => {
+    return async function (dispatch: AppDispatch<QLAction>) {
+        await dispatch(updateChart(preparedChartData, EntryUpdateMode.Publish));
+
+        action();
```
kuzmadom, line 1192: Why is it done this way? Is it possible to call action at the place of use with `then` or `wait`?

src/ui/units/ql/containers/PanePreview/PanePreview.tsx
```diff
@@ -122,11 +139,54 @@ class Preview extends React.PureComponent<PreviewProps, PreviewState> {
...
+                    this.props.openDialogSaveChartConfirm({
+                        onApply,
+                        message: confirmText,
+                    });
+                } else {
+                    resolve(originalAction.apply(this, originalActionArgs));
+                }
+            });
+        };
+    };
```
kuzmadom, line 188: It looks very similar to the code in the Wizard. Is there any way to combine/put into a utility/common component?

src/ui/units/ql/store/actions/ql.ts
stpavlenko: I tried to use `updateChart` action with `await` and looked for more ways, but I didn't find anything, so I decided to put it in a separate action

src/ui/units/ql/containers/PanePreview/PanePreview.tsx
stpavlenko: I think it's possible to put this in a separate function, but it seems to me that it will turn out to be a lot of parameters (confirmText, entryCanBeSaved, onApply, openDialogSaveChartConfirm) and the code will become unreadable. But I'm not sure

kuzmadom (approved):

PR #1577, opened 2024-09-19
Api for working with chart dom elements

kuzmadom:


src/ui/libs/DatalensChartkit/modules/data-provider/charts/ui-sandbox-runtime.ts
```diff
@@ -0,0 +1,202 @@
...
+            throw errorMsg;
+        }
+
+        const value = this.vm.dump(result.value);
+        result.value.dispose();
+
+        return value;
+    }
+
+    private toHandle(value: unknown) {
```
korvin89, line 174: You can add return type `QuickJSHandle` and then you won't need in cast like `this.toHandle(a) as QuickJSHandle`

korvin89 (approved):

PR #1595, opened 2024-09-23
Support copying connections together with copying widget on the dash

mournfulCoroner:


src/ui/units/dash/modules/helpers.ts
```diff
@@ -36,6 +37,9 @@ import {PostMessage} from './postMessage';
 export type CopiedConfigContext = {
     workbookId: WorkbookId;
     fromScope: EntryScope;
+    originalIds?: string[];
```
flops, line 40: Let's use `target` as we made in dashkit

flops (changes requested):

src/ui/utils/copyItems.ts
```diff
@@ -0,0 +1,58 @@
+import type {ConfigItem, ConfigItemData} from '@gravity-ui/dashkit';
+import {DashTabItemType} from 'shared/types';
+import type {ConnectionsData} from 'ui/units/dash/containers/Dialogs/DialogRelations/types';
+
+export const collectWidgetItemIds = (item: {
+    type: ConfigItem['type'];
+    data: ConfigItemData;
+    targetId?: string;
```
flops, line 8: Let's try to use `{targetId: string} | {id: string}` cause in other case we are suggesting that they are both optional

flops (changes requested):

flops (approved):

PR #1629, opened 2024-10-01
Fix column width of the table with virtualization

kuzmadom:


src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/types.ts
```diff
@@ -130,3 +129,25 @@ export type FooterRowViewData = {
...
+    body: {
+        rows: BodyRowViewData[];
+        style?: React.CSSProperties;
+        rowRef?: RowRef;
+    };
+    footer: {
+        rows: FooterRowViewData[];
+        style?: React.CSSProperties;
+    };
+    totalSize: number | undefined;
```
korvin89, line 150: You can skip `undefined`:
```typescript
export type TableViewData = {
    ...
     totalSize?: number;
    ...
};
```

src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/types.ts
```diff
@@ -130,3 +129,25 @@ export type FooterRowViewData = {
...
+    /* rendering table without most options - only to calculate cells size */
+    prerender: boolean;
```
korvin89, line 152: Looks like an optional property

korvin89 (approved):

PR #1674, opened 2024-10-08
Fix endless table rendering cycle

kuzmadom:


src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/BackgroundTable.tsx
```diff
@@ -0,0 +1,82 @@
...
+export const BackgroundTable = React.memo<Props>((props: Props) => {
+    const {
+        dimensions,
+        data: {header, body, footer},
+        onChangeMinWidth,
+    } = props;
+
+    const containerRef = React.useRef<HTMLDivElement | null>(null);
+    const tableRef = React.useRef<HTMLTableElement | null>(null);
+    const tableMinSizes = React.useRef<null | number[]>(null);
```
korvin89, line 33: Why it could not be a `number[]` without `null`?

src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/Table.scss
```diff
@@ -256,6 +256,10 @@
         z-index: -1;
         overflow: auto;
 
+        table {
+            min-width: auto;
```
korvin89, line 260: Does this style make sense?

src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/Table.tsx
```diff
@@ -237,7 +228,7 @@ export const Table = React.memo<Props>((props: Props) => {
                     )}
                     {!noData && (
                         <table
-                            className={b({prepared: !prerender})}
+                            className={b({prepared: true})}
```
korvin89, line 231: Do we actually need this modifier?

src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/cell-width.ts
```diff
@@ -0,0 +1,25 @@
+export function getCellsWidth(args: {
+    cols: {min: number; fixed?: number}[];
+    tableMinWidth: number;
+}): number[] {
+    const {tableMinWidth, cols} = args;
+    const shouldIgnoreFixedWidth = cols.every((col) => col.fixed);
```
korvin89, line 6: You can calculate this value outside this function and use its value as an argument

src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/usePreparedTableData.ts
```diff
@@ -191,12 +192,12 @@ export const usePreparedTableData = (props: {
...
     React.useEffect(() => {
         rowMeasures.current = {};
-    }, [data, cellSizes]);
+    }, [data, dimensions]);
 
     const rowVirtualizer = useVirtualizer({
         count: tableRows.length,
         estimateSize: () => {
-            return 30;
+            return 60;
```
korvin89, line 200: Why exactly 60? :)

kuzmadom: By recommendation at Tanstack Virtual: "If you are dynamically measuring your elements, it's recommended to estimate the largest possible size"
60 - it is approximately row with one line break

src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/Table.tsx
kuzmadom: We definitely need some kind of modifier - tables need different styles (one background for minimum column sizes and one for displaying to the user). But maybe it should be called differently, since the code has undergone changes

src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/Table.scss
kuzmadom: Yes, in the table visible to the user, the minimum width is different - `min-width: 100%;`

src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/BackgroundTable.tsx
kuzmadom: So that we can identify the initial state

src/ui/libs/DatalensChartkit/ChartKit/plugins/Table/renderer/components/Table/cell-width.ts
kuzmadom: I can, but why? It is only needed to calculate the width - it cannot be reused anywhere else

korvin89 (approved):

PR #1679, opened 2024-10-09
Fix EntryScopes

artemipanchuk:


src/ui/components/EntryIcon/EntryIcon.tsx
```diff
@@ -122,24 +122,21 @@ const entityTypeIcons: Record<string, string> = {
...
-        case EntryScope.Report:
-            return 'report';
         case 'broken':
             return 'broken';
         default:
-            return null;
+            // For all new scopes icon type should match scope:
+            // EntryScope.Dataset -> 'dataset'
+            // EntryScope.Folder -> 'folder'
+            return scope;
```
flops, line 139: What about 'unknown' types? Before that `getEntityIconType` was aiming for `null` - now there is no whitelist at all.

artemipanchuk: Normally getEntityIconType should not return null

src/ui/units/workbooks/components/Table/WorkbookEntriesTable/WorkbookEntriesTableTabs.tsx
```diff
@@ -0,0 +1,184 @@
...
+                    workbook={workbook}
+                    onRenameEntry={onRenameEntry}
+                    onDeleteEntry={onDeleteEntry}
+                    onDuplicateEntry={onDuplicateEntry}
+                    onCopyEntry={onCopyEntry}
+                    onShowRelatedClick={onShowRelated}
+                />
+            )}
+            {/* todo: move to other place everything below after update */}
+            {availableScopes?.includes(EntryScope.Report) && (
```
flops, line 160: How are we gonna add `Report` tab without `EntryScope.Report` ?

flops (changes requested):

src/ui/units/workbooks/components/Table/WorkbookEntriesTable/WorkbookEntriesTableTabs.tsx
artemipanchuk: This is customizable component from registry. In next step we will add custom definition of this component to registry in closed source

src/ui/components/EntryIcon/EntryIcon.tsx
flops: https://github.com/datalens-tech/datalens-ui/blob/a4eec11f9ef55e0647da5fdca0797f1a7ebbe710/src/ui/components/EntryIcon/EntryIcon.tsx#L184 - but it does and component that renders this Icon is done to work with `null` value. Current method is working as whitelist filter so that we wouldn't try to get icon for uknown Entry type.

PR #1709, opened 2024-10-17
[WIP]Add treemap visualization

kuzmadom:


src/server/modes/charts/plugins/datalens/preparers/treemap/d3.ts
```diff
@@ -0,0 +1,262 @@
+import type {
```
kuzmadom, line 1: Almost a complete copy of the data preparation for highcharts chart except for the difference in data format and things that didn't make sense

artemipanchuk (approved):

PR #1796, opened 2024-11-11
Chart actions

kuzmadom:


src/shared/types/feature.ts
```diff
@@ -78,6 +78,7 @@ export enum Feature {
     UseMovePermAction = 'UseMovePermAction',
     /** An empty chart for drawing something unusual */
     BlankChart = 'BlankChart',
+    ChartActions = 'ChartActions',
```
korvin89, line 81: Add description please

src/ui/components/Widgets/Chart/components/Content.tsx
```diff
@@ -102,6 +105,40 @@ export const Content = (props: ChartContentProps) => {
     const showContentLoader = showLoader || isExportLoading;
     const showLoaderVeil = veil && !isExportLoading;
 
+    const onAction = React.useCallback(async (actionArgs: {data?: any} = {}) => {
+        const {action, ...args} = actionArgs.data ?? {};
```
korvin89, line 109: I suggest to use `||` here. This code for example will crush execution
```js
// Possible args example (because of data: any)
const actionArgs = {data: false};
// ...
const {action, ...args} = actionArgs.data ?? {};
```

src/ui/components/Widgets/Chart/components/Content.tsx
```diff
@@ -102,6 +105,40 @@ export const Content = (props: ChartContentProps) => {
...
+
+        switch (action) {
+            case 'toast': {
+                const renderMarkdown = await getRenderMarkdownFn();
+                dispatch(
+                    showToast({
+                        ...args,
```
korvin89, line 116: Should we validate user's `args` here?

src/ui/libs/DatalensChartkit/components/Control/Control.tsx
```diff
@@ -152,6 +153,17 @@ class Control<TProviderData> extends React.PureComponent<
         );
     }
 
+    async runAction(args: StringParams) {
+        const {runAction, onAction} = this.props;
```
korvin89, line 157: Could you rename one of `runAction`'s to get rid of variables shadowing? Or you can use inner function in this way `this.props.runAction`

src/ui/libs/DatalensChartkit/modules/constants/constants.ts
```diff
@@ -39,6 +39,7 @@ export type ExportFormatsType = (typeof EXPORT_FORMATS)[keyof typeof EXPORT_FORM
 export const CLICK_ACTION_TYPE = {
     SET_PARAMS: 'setParams',
     SET_INITIAL_PARAMS: 'setInitialParams',
+    USE_ACTION: 'useAction',
```
korvin89, line 42: Maybe `dispatch`? `use` prefix used by React hooks usually

src/ui/units/editor/configs/chart/chart-editor-types.ts
```diff
@@ -640,6 +640,15 @@ export function getChartEditorTypes(type: string) {
         },
     } as Record<string, ChartEditorType>;
 
+    if (Utils.isEnabledFeature(Feature.ChartActions)) {
+        chartEditorTypes[EDITOR_TYPE.CONTROL_NODE].tabs.push({
+            name: 'Actions',
```
korvin89, line 645: If a user wants to show more than one toast, they will be prevented from doing so by the same values of the name field. Is this expected?

korvin89 (approved):

melikhov-dev (approved):

PR #1829, opened 2024-11-21
Move to navigate route by default

goshander:


src/ui/units/dash/modules/helpers.ts
```diff
@@ -176,7 +176,7 @@ export function addOperationForValue(args: {
 
 export function getChartEditLink(entryId: string) {
     return Utils.isEnabledFeature(Feature.UseNavigation)
-        ? `${window.DL.endpoints.charts}/navigation/${entryId}`
+        ? `${window.DL.endpoints.charts}/navigate/${entryId}`
```
jhoncool, line 179: I think here should be `/navigate/${entryId}` or `${window.location.origin}/navigate/${entryId}`

goshander: Yep, I fix it a second ago 🙏

/navigation - for old behaviour with enabled UseNavigation flag
/navigate - for new route wuth support not only for wizard

jhoncool: Are you sure that we need window.DL.endpoints.charts for new route?

src/ui/units/dash/modules/helpers.ts
```diff
@@ -177,7 +177,7 @@ export function addOperationForValue(args: {
 export function getChartEditLink(entryId: string) {
     return Utils.isEnabledFeature(Feature.UseNavigation)
         ? `${window.DL.endpoints.charts}/navigation/${entryId}`
-        : `${window.DL.endpoints.wizard}/${entryId}`;
+        : `${window.DL.endpoints.charts}/navigate/${entryId}`;
```
jhoncool, line 180:
```suggestion
        : `${window.location.origin}/navigate/${entryId}`;
```

goshander:
@jhoncool change to simple `/navigate`

src/ui/units/dash/modules/helpers.ts
goshander: No, you are right

src/ui/libs/DatalensChartkit/modules/menu/menu.ts
```diff
@@ -58,9 +58,6 @@ export const getChartkitMenuByType = (props?: GetChartkitMenuByType) => {
             },
             [MenuItemsIds.EDIT]: {
                 title: i18n('dash.chartkit-menu.view', 'button_edit'),
-                action: ({propsData: {id, params}}) => {
```
jhoncool, line 61: Why don't we need it?

goshander: It is override default behaviour, but probably we can delete it later

jhoncool (approved):

PR #1833, opened 2024-11-22
Fix filtration by action params

DaryaLari:


src/ui/libs/DatalensChartkit/ChartKit/helpers/action-params-handlers.ts
```diff
@@ -84,16 +85,19 @@ export function addParams(params: StringParams, addition: StringParams = {}) {
...
         }
 
-        if (!Array.isArray(acc[key])) {
-            acc[key] = [acc[key] as string];
-        }
+        let currentKeyValue: string[];
 
-        if ((acc[key] as string[]).every(isEmptyParam)) {
-            acc[key] = [];
+        if (!Array.isArray(acc[key])) {
```
kuzmadom, line 90: Optional: it is better to invert the condition so that it does not start with a negative statement

kuzmadom (approved):

kuzmadom (approved):

PR #1840, opened 2024-11-25
Undo/redo in datasets

korvin89:


src/ui/units/datasets/containers/Dataset/useHistoryActions.tsx
```diff
@@ -0,0 +1,96 @@
...
+                view: 'flat',
+                disabled: !canGoForward,
+                title: i18n('component.action-panel.view', 'button_redo'),
+                hotkey: REDO_HOTKEY.join('+'),
+            },
+        ];
+    }, [canGoBack, canGoForward, enabled, handleGoBack, handleGoForward]);
+    const actions = useAdditionalItems({items});
+
+    useBindHotkey({key: UNDO_HOTKEY, handler: handleGoBack, options});
```
artemipanchuk, line 92: Now buttons are shown only if flag is true, but hotkeys are always bound. Hotkeys should be bound only if flag is true, too.

src/ui/store/actions/editHistory.ts
```diff
@@ -127,7 +137,7 @@ export function addEditHistoryPoint({unitId, newState}: {unitId: string; newStat
                 diff = jdp.diff(oldState, newState);
             }
 
-            dispatch(_addEditHistoryPoint({unitId, diff, state: newState}));
+            dispatch(_addEditHistoryPoint({unitId, diff, state: newState, stacked}));
```
artemipanchuk, line 140: Can you explain why do we need stacked?

korvin89: I've tried to explain it [here](https://github.com/datalens-tech/datalens-ui/pull/1840/files#diff-c96d83998e87ddd3090121dd876da4db5e6ade7a8d782505b3cafcd3ac85d570R111-R114). Maybe this comment isn't clear

artemipanchuk (approved):

PR #1846, opened 2024-11-27
Add collections e2e tests

stepanenkoxx:


scripts/e2e/us-dump.sh
```diff
@@ -40,8 +40,9 @@ if [ $CONNECTION_CHECK_RESULT_STATUS_CODE -ne 0 ]; then
     exit 1
 fi
 
-DUMP_CMD="pg_dump --dbname=$US_CONNECTION --column-inserts -a -t workbooks -t entries -t revisions -t links"
-docker exec -it $US_CONTAINER_ID /bin/sh -c "$DUMP_CMD" | grep -Ev "^(--|SET|SELECT pg_catalog.set_config)" >$DUMP_TMP
+DUMP_CMD="pg_dump --dbname=$US_CONNECTION --disable-triggers --column-inserts -a -t collections -t workbooks -t entries -t revisions -t links"
+docker exec -it $US_CONTAINER_ID /bin/sh -c "$DUMP_CMD" |
+tee >(grep -Ev "^(--|SET|SELECT pg_catalog.set_config|pg_dump:)" >$DUMP_TMP) | grep "pg_dump:"
```
stepanenkoxx, line 45: After adding the collections table to the dump, I started getting pg_dump warnings in the dump file due to a circular foreign key in the collections table.

pg_dump: warning: there are circular foreign-key constraints on this table:
pg_dump: detail: collections
pg_dump: hint: You might not be able to restore the dump without using --disable-triggers or temporarily dropping the constraints.
pg_dump: hint: Consider using a full dump instead of a --data-only dump to avoid this problem.
I added ```--disable-triggers``` flag to pg_dump to solve problem if we will have collections with circular dependencies in the future
But warnings were still written to the file
To resolve this, I filtered the warnings from the dump file and redirected the warnings output to the console.

stepanenkoxx, line 45: After adding the ```collections``` table to the dump, I started getting pg_dump warnings in the dump file due to a circular foreign key in the ```collections``` table

```
pg_dump: warning: there are circular foreign-key constraints on this table:
pg_dump: detail: collections
pg_dump: hint: You might not be able to restore the dump without using --disable-triggers or temporarily dropping the constraints.
pg_dump: hint: Consider using a full dump instead of a --data-only dump to avoid this problem.
```
I added ```--disable-triggers``` flag to pg_dump to solve problem if we will have collections with circular dependencies in the future
But warnings were still written to the file
To resolve this, I filtered the warnings from the dump file and redirected the warnings output to the console.

tests/opensource-suites/collections/screenshots/collectionsPage.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import {CollectionsPageQa} from '../../../../src/shared';
+import {openTestPage, slct} from '../../../utils';
+import datalensTest from '../../../utils/playwright/globalTestDefinition';
+
+datalensTest.describe('Collections page', () => {
+    datalensTest(
+        'Collections page displays list of collections and workbooks @screenshot',
+        async ({page}: {page: Page}) => {
+            await page.clock.setFixedTime(new Date('2025-01-01T10:00:00'));
```
kuzmadom, line 11: Since the collection contains an entity that is used in other tests and is being changed (e2e workbook), a fixed time will not help much in this case. I think it would be cleaner to create a separate collection with a new (possibly empty) workbook

tests/opensource-suites/collections/screenshots/collectionsPage.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+            await openTestPage(page, `/collections`);
+
+            const pageContainer = page.locator(slct(CollectionsPageQa.PageContainer));
+
+            await page.setViewportSize({width: 1200, height: 500});
```
kuzmadom, line 17: Why was it necessary to set the size?

tests/opensource-suites/collections/screenshots/collectionsPage.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+            await expect(pageContainer).toHaveScreenshot();
```
kuzmadom, line 19: If the test checks the general state of the page (and not the layout), then it is more useful to check for errors in the console, on the network, etc. instead of a screenshot

tests/opensource-suites/collections/screenshots/collectionsPage.test.ts
stepanenkoxx: I'm fixing time here to make stable this part of screenshot
"last modified" there depends on current date

Screenshot 2024-11-28 at 15 02 56

tests/opensource-suites/collections/screenshots/collectionsPage.test.ts
stepanenkoxx: with default height the bottom of the screenshot remains empty
i reduced the height to make the screenshot file weight smaller

tests/opensource-suites/collections/screenshots/collectionsPage.test.ts
stepanenkoxx: removed the screenshot test :)

tests/opensource-suites/collections/contentTableItemsLinks.test.ts
```diff
@@ -0,0 +1,28 @@
...
+
+            await page.waitForURL(WorkbooksUrls.E2EWorkbook);
+        });
+
+        datalensTest('Collection row link leads to collection page', async ({page}) => {
+            const collectionsPage = new CollectionsPage({page});
+
+            await collectionsPage.clickCollectionRowLink();
+
+            await page.waitForURL(CollectionsUrls.E2ECollection);
```
kuzmadom, line 25: If the purpose of the test was to check the formation of correct links, then it is probably fine. 
But if it was necessary to check the transition to the page, then it is not enough to check the url - you need to check some other sign that the expected page has been rendered.

kuzmadom (dismissed):

Marginy605 (approved):

tests/opensource-suites/collections/contentTableItemsLinks.test.ts
stepanenkoxx: added some pages content visibility checks

kuzmadom (approved):

PR #1883, opened 2024-12-05
Support wrappedHtml in Table cell formattedValue

korvin89:


src/ui/libs/DatalensChartkit/ChartKit/components/Widget/components/Table/Bar/Bar.tsx
```diff
@@ -38,7 +40,15 @@ export const Bar = (props: BarProps) => {
         showSeparator,
     });
 
-    const displayedValue = typeof formattedValue === 'string' ? formattedValue : value;
+    let displayedValue: React.ReactNode;
+
+    if (isWrappedHTML(formattedValue)) {
```
kuzmadom, line 45: Shouldn't I also check the `value` here? not only `formattedValue`

src/ui/libs/DatalensChartkit/ChartKit/components/Widget/components/Table/WrappedHTMLNode.tsx
```diff
@@ -10,6 +11,14 @@ type Props<T extends React.ElementType> = {
     className?: string;
 };
 
+export function isWrappedHTML(value: unknown): value is WrappedHTML {
+    if (!value || typeof value !== 'object') {
+        return false;
+    }
+
+    return Object.keys(value).length === 1 && WRAPPED_HTML_KEY in value;
```
kuzmadom, line 19: Is it possible that checking for the number of keys is unnecessary? I can imagine that in the future there will be an additional parameter for generating html, for example

kuzmadom (approved):

src/ui/libs/DatalensChartkit/ChartKit/components/Widget/components/Table/WrappedHTMLNode.tsx
korvin89: Sure, lets remove this condition

kuzmadom (approved):

PR #1897, opened 2024-12-06
Fix ui-sandbox error

kuzmadom:


korvin89 (dismissed):

src/ui/libs/DatalensChartkit/modules/data-provider/charts/ui-sandbox.ts
```diff
@@ -331,21 +325,46 @@ async function getUnwrappedFunction(args: {
...
 
-        const performance = Performance.getDuration(runId);
-        if (options?.totalTimeLimit) {
-            options.totalTimeLimit = Math.max(0, options.totalTimeLimit - Number(performance));
-        }
+            return unwrapHtml(result, parseHtml);
+        } catch (e) {
+            const performance = Performance.getDuration(runId);
+            if (performance && e?.message === 'interrupted') {
+                if (options?.totalTimeLimit && performance > options?.totalTimeLimit) {
```
kuzmadom, line 346: @melikhov-dev 
The error fix is here. Previously, this code was called at the beginning - before executing next  function. If the total limit was exceeded as a result of the execution of this function, then the user was likely to receive an interupped error (the remainder of the total limit was less than a single function exec limit). We must first check the total limit and only then check if it does not exceed the single function exec  limit

melikhov-dev (approved):

PR #1937, opened 2024-12-12
Add nullMode to GradientColorConfig & Up @gravity-ui/uikit 6.37.0 -> 6.39.0

DaryaLari:


src/ui/units/wizard/components/Dialogs/DialogColor/DialogColor.tsx
```diff
@@ -191,10 +194,11 @@ class DialogColorComponent extends React.Component<Props, State> {
                 polygonBorders,
                 reversed,
                 thresholdsMode,
-                leftThreshold,
-                middleThreshold,
-                rightThreshold,
+                leftThreshold: leftThreshold ? leftThreshold : undefined,
+                middleThreshold: middleThreshold ? middleThreshold : undefined,
+                rightThreshold: rightThreshold ? rightThreshold : undefined,
```
kuzmadom, line 199: As far as I understand it, it works now because the values are set in strings, right? But it is better to explicitly check for undefined if this is the goal - in this case, 0 will not be lost when the values become numbers (as it should be)

src/ui/units/wizard/components/Dialogs/DialogColor/DialogColorGradient/DialogColorGradient.tsx
```diff
@@ -30,6 +33,11 @@ export const AUTO = 'auto';
 const MANUAL = 'manual';
 const TOOLTIP_DELAY_CLOSING = 250;
 
+const NULLS_OPTIONS = [
+    {value: 'ignore', content: i18n('wizard', 'label_painting-ignore')},
+    {value: 'as-0', content: i18n('wizard', 'label_painting-as-0')},
```
kuzmadom, line 38: You can use GradientNullModes values instead of strings

src/ui/units/wizard/components/Dialogs/DialogColor/DialogColorGradient/DialogColorGradient.tsx
```diff
@@ -75,11 +84,17 @@ class DialogColorGradientBody extends React.Component<Props> {
...
-        const {extra = {}, item, gradients} = this.props;
+        const {extra = {}, item, gradients, visualizationId} = this.props;
+
+        const nullModeValue = this.props.gradientState.nullMode ?? DEFAULT_NULL_MODE;
 
         const currentPalette = this.getCurrentGradientPalette(gradients);
         const options = getGradientSelectorItems(gradients);
 
+        const canSetNullMode = (
+            ALLOWED_FOR_NULL_MODE_VISUALIZATIONS as (string | undefined)[]
```
kuzmadom, line 95: It seems that you should not cast to `undefined` here

src/ui/units/wizard/constants/dialogColor.ts
```diff
@@ -25,4 +26,10 @@ export const DEFAULT_GRADIENT_STATE = {
     middleThreshold: undefined,
     rightThreshold: undefined,
     gradientPalette: DEFAULT_TWO_POINT_GRADIENT,
+    nullMode: undefined,
```
kuzmadom, line 29: why is it not equal to an DEFAULT_NULL_MODE here?

src/ui/units/wizard/reducers/utils/clearUnusedVisualizationItems.ts
```diff
@@ -1,4 +1,4 @@
-import {pick} from 'lodash';
+import {omit, pick} from 'lodash';
```
kuzmadom, line 1: better
```
import omit from 'lodash/omit';
import pick from 'lodash/pick';
```

src/ui/units/wizard/selectors/dialogColor.ts
```diff
@@ -17,6 +17,9 @@ export const selectDialogColorPaletteState = (state: DatalensGlobalState) =>
 export const selectDialogColorGradientState = (state: DatalensGlobalState) =>
     state.wizard.dialogColor.gradientState;
 
+export const selectDialogColorVisualizationId = (state: DatalensGlobalState) =>
+    state.wizard.visualization.visualization?.id;
```
kuzmadom, line 21: It seems that this does not apply directly to the color settings dialog? It is better to put it in a common place

src/ui/units/wizard/components/Dialogs/DialogColor/DialogColorGradient/DialogColorGradient.tsx
DaryaLari: fixed

src/ui/units/wizard/constants/dialogColor.ts
DaryaLari: nullMode should be undefined by default, because it is applicable not for all types of charts

src/ui/units/wizard/selectors/dialogColor.ts
DaryaLari: Removed this selector and replaced its usage with another existing one

kuzmadom (approved):

PR #1948, opened 2024-12-17
Add blank chart events and state

kuzmadom:


src/ui/libs/DatalensChartkit/ChartKit/plugins/BlankChart/renderer/BlankChartWidget.tsx
```diff
@@ -35,19 +39,62 @@ const BlankChartWidget = (props: BlankChartWidgetProps) => {
...
+                if (!options?.silent) {
+                    render();
                 }
+            },
+        });
+
+        return () => {
+            chartStorage.delete(generatedId);
+        };
+    }, [generatedId, dimensions]);
```
korvin89, line 60: Do we actually need `dimensions` here?

korvin89 (approved):

src/ui/libs/DatalensChartkit/ChartKit/plugins/BlankChart/renderer/BlankChartWidget.tsx
kuzmadom: Yes, we need to get the current state when changing the dimensions

PR #1951, opened 2024-12-17
Add chart widget title hint (on dash)

Marginy605:


src/ui/components/Widgets/Chart/components/WidgetHeader.tsx
```diff
@@ -61,17 +64,49 @@ export const WidgetHeader = (props: HeaderProps) => {
...
+                    <div className={b('mobile-title-wrap')}>
+                        {widgetTitle}
+                        {Boolean(widgetTitleHint) && (
+                            <MarkdownHelpPopover
+                                markdown={widgetTitleHint}
+                                className={b('chart-title-hint')}
+                                buttonProps={{
+                                    className: b('chart-title-hint-button'),
+                                }}
+                                {...(DL.IS_MOBILE
```
mournfulCoroner, line 89: There's `if (DL.IS_MOBILE &&` in the if block just above, so the second check of it is not necessary

mournfulCoroner (approved):

PR #1967, opened 2024-12-20
Add check of connections for publication preview

mournfulCoroner:


src/shared/schema/mix/helpers/entries.ts
```diff
@@ -1,18 +1,26 @@
...
+import type {
+    CheckConnectionsForPublicationResponse,
+    CheckDatasetsForPublicationResponse,
+} from '../../types';
 import type {EntryFields} from '../../us/types';
 
-export function filterDatasetsIdsForCheck(entries: Pick<EntryFields, 'entryId' | 'scope'>[]) {
-    return entries
-        .filter((entry) => {
-            if (!('scope' in entry)) {
```
jhoncool, line 10: Сan you not delete this check? Of course it would be better to additionally request entries in switchPublicationStatus with scopes

jhoncool (approved):

PR #1968, opened 2024-12-23
Fix opening of dataset entries with "new" in the end of name

korvin89:


src/ui/units/datasets/helpers/utils.ts
```diff
@@ -150,6 +150,7 @@ export default class DatasetUtils {
     }
 }
 
-export function isCreationProcess() {
-    return /new$/.test(window.location.pathname);
+export function isCreationProcess(pathname = '') {
+    const lastPathnamePart = pathname.split('/').filter(Boolean).slice(-1)[0];
```
kuzmadom, line 154: It would probably be possible to use a matchPath if you need to check one specific route

kuzmadom (approved):

PR #1974, opened 2024-12-24
Add headers for dataset request

melikhov-dev:


src/server/components/charts-engine/components/processor/data-fetcher.ts
```diff
@@ -230,14 +230,14 @@ export class DataFetcher {
                     cookie: req.headers.cookie,
                 },
             };
+            const datasetHeaders = req.headers;
```
kuzmadom, line 233: Why **dataset**Headers? There may be requests not only from datasets, but also from other sources, no?

kuzmadom (approved):

kuzmadom (approved):

PR #2041, opened 2025-01-22
Validate url schema

kuzmadom:


korvin89 (commented):
~~I suppose we can remove `validateUrl` from our code base?~~

korvin89 (approved):

PR #2090, opened 2025-01-31
Use separate isEnabledFeature function

mournfulCoroner:


src/ui/utils/isEnabledFeature.ts
```diff
@@ -0,0 +1,12 @@
+import _get from 'lodash/get';
```
melikhov-dev, line 1: lets rename to `get`?

src/ui/utils/isEnabledFeature.ts
```diff
@@ -0,0 +1,12 @@
...
+import {DL} from 'ui/constants/common';
+
+export const isEnabledFeature = (featureName: string) => {
+    const featureDynamicStatus = _get(DL.DYNAMIC_FEATURES, featureName);
+
+    if (typeof featureDynamicStatus !== 'undefined') {
```
melikhov-dev, line 7: I think we can use `featureDynamicStatus !== undefined`

src/ui/utils/isEnabledFeature.ts
```diff
@@ -0,0 +1,12 @@
...
+        return featureDynamicStatus;
```
melikhov-dev, line 8: Boolean(featureDynamicStatus) ?

src/ui/utils/isEnabledFeature.ts
```diff
@@ -0,0 +1,12 @@
+import _get from 'lodash/get';
+import {DL} from 'ui/constants/common';
+
+export const isEnabledFeature = (featureName: string) => {
```
melikhov-dev, line 4:
```suggestion
export const isEnabledFeature = (featureName: string): Boolean => {
```

src/ui/utils/isEnabledFeature.ts
mournfulCoroner: I changed the code a bit

PR #2104, opened 2025-02-05
Add auth api to gateway, authenticated user to DL

jhoncool:
Add auth api to gateway
Add authenticated user to DL, layout


src/shared/schema/auth/types/users.ts
```diff
@@ -0,0 +1,92 @@
...
+export interface DeleteUserArgs {
+    userId: string;
+}
+
+export interface GetUserProfileArgs {
+    userId: string;
+}
+
+export interface GetUserProfileResponse {
+    profile: {
```
DaryaLari, line 42: Maybe it is better to move profile data to a separate interface and reuse it in GetUsersListResponse?

jhoncool: It may change in the future, So for me better wouldn't mix it.

DaryaLari (approved):

mournfulCoroner (approved):

PR #2148, opened 2025-02-18
Add validation in sign in form

mournfulCoroner:


DaryaLari:
It would be also nice to have trimming of values from whitespaces before submiting

DaryaLari (approved):

PR #2151, opened 2025-02-18
Edit user role dialog + improve actions handling

DaryaLari:


src/ui/units/service-settings/components/UsersList/UsersList.tsx
```diff
@@ -170,6 +206,23 @@ const UsersList = () => {
...
+                    />
+                )}
+                {deleteUserDialogOpenForUser && (
+                    <DeleteUserDialog
+                        open
+                        onClose={() => setDeleteUserDialogOpenForUser(undefined)}
+                        onSuccess={handleUserDeleteSuccess}
+                        userId={deleteUserDialogOpenForUser.userId}
+                    />
+                )}
```
mournfulCoroner, line 225: Why is the modal display controlled not only through the open prop, as is the case with the profile?
Just in case: if you use a condition other than open, the animation of the dialog is lost. (but this is not so critical, because we have quite a lot of dialogues with lost animations)

src/ui/units/service-settings/components/UsersList/UsersList.tsx
```diff
@@ -94,9 +99,20 @@ const UsersList = () => {
...
+        ListUser | undefined
+    >();
+
     React.useEffect(() => {
         dispatch(resetServiceUsersList());
         dispatch(getUsersList({pageSize: USERS_PAGE_SIZE}));
+
+        return () => {
+            dispatch(resetServiceUsersList());
+        };
```
mournfulCoroner, line 115: Maybe then it's worth removing the reset at the beginning of the function?

src/ui/units/service-settings/components/UsersList/UsersList.tsx
```diff
@@ -141,23 +157,43 @@ const UsersList = () => {
...
+
+    const handleUserRoleChangeSuccess = React.useCallback(() => {
+        dispatch(resetServiceUsersList());
+        dispatch(
+            getUsersList({
+                pageSize: USERS_PAGE_SIZE,
+                ...filters,
+            }),
+        );
+    }, [dispatch, filters]);
```
mournfulCoroner, line 178: It can be made into a single resetTable function

src/ui/units/service-settings/components/UsersList/UsersList.tsx
DaryaLari: I think, it's better to clear data after leaving page, so that it wouldn't take up browser memory space when the data is not more needed

mournfulCoroner (approved):

PR #2167, opened 2025-02-20
Bump weblate-action

dgaponov:


mournfulCoroner (approved):

mournfulCoroner:
The changes do not affect the code, they only affect the i18n actions, so passing the tests is not required.

PR #2169, opened 2025-02-21
Fix en translations

mournfulCoroner:


mournfulCoroner:
It's PR to fix the i18n flow. It doesn't contain code changes, only keysets, so tests are not necessary.

Marginy605 (approved):

PR #2227, opened 2025-02-28
Add generate password button

mournfulCoroner:


src/ui/units/auth/components/formControls/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+function getRandomSymbolFromString(set: string) {
+    const randomNumbers = new Uint32Array(1);
+    window.crypto.getRandomValues(randomNumbers);
+    const index = randomNumbers[0] % set.length;
+    return set[index];
+}
+
+export function generateRandomPassword() {
```
jhoncool, line 22: @mournfulCoroner can you add some unit test?

PR #2240, opened 2025-03-05
Up weblate action version to 1.49.0

Marginy605:


mournfulCoroner (approved):

Marginy605:
Affects only on i18n-check action with changed keysets. No need to wait all tests and etc checks

PR #2257, opened 2025-03-07
Make dash data optional if possible

flops:


src/server/modes/charts/plugins/data-api-json-schema.ts
```diff
@@ -0,0 +1,633 @@
...
+                                minLength: 1,
+                            },
+                            namespace: {
+                                const: DASH_DEFAULT_NAMESPACE,
+                                // uncomment when namespace is different from default
+                                // type: 'string',
+                                // minLength: 1,
+                            },
+                            sourceType: {
+                                enum: Object.values(DashTabItemControlSourceType),
```
mournfulCoroner, line 305: there must be only manual and dataset source types.

mournfulCoroner (dismissed):

PR #2269, opened 2025-03-10
Support RLSv2 config in datasets

korvin89:


src/ui/units/datasets/store/selectors/dataset.ts
```diff
@@ -31,7 +33,13 @@ export const sourcesSelector = (state: DatalensGlobalState) => state.dataset.con
 export const avatarsSelector = (state: DatalensGlobalState) => state.dataset.content.source_avatars;
 export const relationsSelector = (state: DatalensGlobalState) =>
     state.dataset.content.avatar_relations;
-export const rlsSelector = (state: DatalensGlobalState) => state.dataset.content.rls;
+export const rlsSelector = (state: DatalensGlobalState) => {
+    if (isEnabledFeature(Feature.EnableRLSV2)) {
+        return state.dataset.content.rls || [];
```
DaryaLari, line 38: Shouldn't it be `rls2` here?

src/ui/units/datasets/containers/DatasetEditor/DatasetEditor.js
```diff
@@ -297,7 +299,8 @@ class DatasetEditor extends React.Component {
         const {guid} = field;
 
         if (rls && guid) {
-            const rlsField = rls[guid] || '';
+            const rlsFieldDefault = isEnabledFeature(Feature.EnableRLSV2) ? [] : '';
```
DaryaLari, line 302: Maybe it would be better to define this constant outside of the component or as static field? And then replace all similar conditions in this component with new constant usage?

src/ui/units/datasets/store/selectors/dataset.ts
korvin89: Nope

src/ui/units/datasets/containers/DatasetEditor/DatasetEditor.js
korvin89: I think it's more transparent to have such things directly in the place where they are applied, it's not a constant.

src/ui/units/datasets/store/selectors/dataset.ts
korvin89: Oh, yep, I'll fix it

DaryaLari (approved):

PR #2299, opened 2025-03-19
Fix workbookSelectionMode in CollectionStructureDialog

imsitnikov:
In the future, need to separate collections selection and workbook selection in the dialog...
This logic becomes too confusing.

### DIFF
--- a/src/ui/components/CollectionsStructure/CollectionStructureDialog/CollectionStructureDialog.tsx
+++ b/src/ui/components/CollectionsStructure/CollectionStructureDialog/CollectionStructureDialog.tsx
@@ -12,6 +12,7 @@ import type {
     GetStructureItemsResponse,
     OrderBasicField,
     OrderDirection,
+    WorkbookWithPermissions,
 } from 'shared/schema';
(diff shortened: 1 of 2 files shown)


src/ui/components/CollectionsStructure/CollectionStructureDialog/CollectionStructureDialog.tsx
```diff
@@ -155,24 +156,29 @@ export const CollectionStructureDialog = React.memo<Props>(
...
-                            ...args,
-                            page: curPage,
-                        });
-                    } else {
-                        return result;
-                    }
-                });
+                return dispatch(getStructureItems({...args, includePermissionsInfo})).then(
+                    (result) => {
+                        if (result?.items.length === 0 && result.nextPageToken !== null) {
```
Sergey-weber, line 167: it was already here, I just wanted to know in which case we might have an empty list and not empty the nextPageToken so that we would like to call recursion?

imsitnikov: In case we don't have access to all entries on the current page.

Sergey-weber (approved):

PR #2354, opened 2025-04-06
[DRAFT] Workbook transfer api

stepanenkoxx:


src/server/controllers/workbook-transfer.ts
```diff
@@ -1,15 +1,50 @@
...
+
+type ImportRequestBody = {
+    data: object;
+    workbookId: string;
+    idMapping: TransferIdMapping;
+    scope: EntryScope;
+};
+
+type ImportSuccessResponseBody = {
+    entryId?: string;
```
stepanenkoxx, line 36: I guess it could be undefined, cause we can't create entry when getting critical notification

src/server/controllers/workbook-transfer.ts
```diff
@@ -1,15 +1,50 @@
...
 import type {DatalensGatewaySchemas} from '../types/gateway';
 import type {GatewayApiErrorResponse} from '../utils/gateway';
 
-export const workbooksExportController = {
-    export: async (req: Request, res: Response) => {
+type ExportRequestBody = {
+    entryId: string;
+    idMapping: TransferIdMapping;
+    workbookId: string;
+};
```
stepanenkoxx, line 15: scope also could be sent here if needed

PR #2365, opened 2025-04-08
Fix allowed tags

kuzmadom:


src/ui/libs/DatalensChartkit/modules/html-generator/constants.ts
```diff
@@ -18,7 +18,7 @@ export const ALLOWED_TAGS = [
     'button',
     'caption',
     'circle',
-    'clipPath',
+    'clippath',
```
Marginy605, line 21: Now it's inconvenient to keep actual list of tags and attrs in docs :/

Marginy605:
By the way, could you check and add this attrs, please?
- xmlns:xlink
- overflow

Marginy605 (changes requested):

PR #2404, opened 2025-04-17
Add i18n-keysets-sorting-check gh workflow

korvin89:


korvin89:
@Marginy605 could u approve weblate translations? Looks like a falsy triggering because of the refactoring existing keys

Marginy605:
> @Marginy605 could u approve weblate translations? Looks like a falsy triggering because of the refactoring existing keys

I suppose that it's not falsy but because of wrong order of keysets was merge to main

image

I'll approve it

Marginy605:
@korvin89 ✅ done

scripts/i18n-keysets-clear copy.js
```diff
@@ -0,0 +1,13 @@
+/**
```
Marginy605, line 1: looks like it's better to change this filename

Marginy605 (changes requested):

scripts/i18n-keysets-clear copy.js
korvin89: There is already such file

Marginy605 (approved):

PR #2429, opened 2025-04-24
Support resolve authenticated users logins in the change history and entity list

mournfulCoroner:


src/ui/units/auth/store/selectors/usersByIds.ts
```diff
@@ -0,0 +1,41 @@
...
+            preparedUserById = {
+                status: 'loading',
+            };
+        } else if (user.error === null && user.data) {
+            const data = user.data;
+            let login: string;
+            if (data.source === 'system') {
+                login = data.subject.login;
+            } else {
+                login = data.subject.login || getUserDisplayName(data.subject, true);
```
mournfulCoroner, line 23: for some reason, `login` can be null in the types, so I added a fallback to the first/last name. I'm not sure if it's really necessary here

jhoncool (approved):

DaryaLari (approved):

PR #2504, opened 2025-05-20
Add docs links and fix some bugs for export/import workbooks

mournfulCoroner:
- Add two docs links 
- If there are critical notifications, only they are displayed on the screen.
- When uploading the correct file, its name is inserted in the Workbook Name field.
- When using Escape, you can now close the export correctly.
- Fixed endless loading with broken JSON in a file.
- Added a fallback in case the workflow was deleted after a day


src/ui/units/collections/components/CollectionPage/CollectionPage.tsx
```diff
@@ -76,6 +76,10 @@ export const CollectionPage = () => {
         [curCollectionId, dispatch, history],
     );
 
+    const handleCreateDialogAction = React.useCallback(() => {
+        handleOpenCreateDialog();
+    }, [handleOpenCreateDialog]);
```
kuzmadom, line 81: Maybe we don't need a wrapper here? It doesn't do anything.

PR #2534, opened 2025-05-26
Fix loader for external selectors

flops:


Marginy605 (approved):

Marginy605:
Unfortunately flag GroupControlsLoadingLock causes race in tests' cases and breaks them

PR #2627, opened 2025-06-27
Fix terms placement in fixed header

DaryaLari:


src/ui/units/dash/containers/FixedHeader/FixedHeader.tsx
```diff
@@ -200,6 +203,36 @@ export function FixedHeaderWrapper({
         };
     }, [wrapperRef, topOffset]);
 
+    React.useEffect(() => {
+        const observer = new ResizeObserver(([el]) => {
+            if (el && scrollableContainerRef.current) {
+                const {height} = el.contentRect;
+                const maxHeightPx = getComputedStyle(scrollableContainerRef.current).maxHeight;
```
flops, line 210: `getComputedStyle` - is really heavy to use on resize. Maybe fix root cause, make .dfn-term hidden while it's not open and use only clientHeight and scrollHeight to determine when overflow is needed.
And you can use el.target as reference for domNode

flops (changes requested):

flops (approved):

PR #2640, opened 2025-07-04
Add more qa and utils to workbooks

mournfulCoroner:


tests/page-objects/BasePage.ts
```diff
@@ -87,4 +87,27 @@ export class BasePage {
...
         return extractEntryId(entryId);
     }
+
+    async checkForFailedRequests(waitingCallback: () => Promise<void>) {
+        // check that the requests are completed successfully before the callback is resolved.
+        const failedRequests: Response[] = [];
+
+        const onResponse = (response: Response) => {
+            const status = response.status();
+            if (status >= 400) {
```
korvin89, line 97: Would it be more reliable to rely on some kind of boolean flag that indicates that the request has failed?

korvin89 (dismissed):

PR #2675, opened 2025-07-14
Disallow dragging parameters to the Y section

kuzmadom:


src/ui/units/wizard/utils/placeholder.ts
```diff
@@ -69,3 +83,43 @@ export function getPlaceholderAxisModeMap(args: {placeholder: Placeholder; axisM
...
+        case WizardVisualizationId.Area:
+        case WizardVisualizationId.Area100p:
+        case WizardVisualizationId.Column:
+        case WizardVisualizationId.Column100p: {
+            if (!isNumberField(field) && !isDateField(field)) {
+                return ![PlaceholderId.Y, PlaceholderId.Y2].includes(
+                    placeholderId as PlaceholderId,
+                );
+            }
+            break;
```
DaryaLari, line 113: It is a bit hard to read these negative conditions. It would be better to replace them

PR #2691, opened 2025-07-23
Fix yfm styles

kuzmadom:


Marginy605 (approved):

kuzmadom:
Statoscope check - an incorrect error when changing styles, in fact everything is ok

PR #2773, opened 2025-08-13
Hide axis type in placeholder dialog for charts with yagr visualization

dvortsovs:


src/ui/units/wizard/components/Dialogs/DialogPlaceholder/utils.ts
```diff
@@ -9,6 +13,22 @@ export function isAxisTypeEnabled(visualizationId: WizardVisualizationId) {
...
+) {
+    return (
+        qlChartType === QLChartType.Monitoringql &&
+        [
+            WizardVisualizationId.Line,
+            WizardVisualizationId.Area,
+            WizardVisualizationId.Area100p,
+            WizardVisualizationId.Column,
+            WizardVisualizationId.Column100p,
+        ].includes(visualizationId)
```
kuzmadom, line 28: Should use isYAGRVisualization from https://github.com/datalens-tech/datalens-ui/blob/feb556bb13d06b29c2e2ba52c640301e689d0ec7/src/shared/utils/visualization-check.ts#L9

kuzmadom (approved):

PR #2778, opened 2025-08-14
Add FieldSettings dialog to dataset

kuzmadom:


korvin89 (approved):
Size `m` is default for `RadioButton`, `TextInput` components

PR #2789, opened 2025-08-19
Add entry description dialog usage

Alutery:


src/ui/components/DialogEntryDescription/DialogEntryDescription.tsx
```diff
@@ -152,12 +148,6 @@ export const DialogEntryDescription: React.FC<DialogEntryDescriptionProps> = (pr
                                 {i18n('button_edit')}
                             </Button>
                         )}
-                        {!loading && (
```
Alutery, line 155: Forgot to remove in the previous PR

PR #2792, opened 2025-08-20
Add axis labels formatting for highcharts

dvortsovs:


src/shared/types/wizard/placeholder.ts
```diff
@@ -21,6 +21,8 @@ export interface PlaceholderSettings {
     holidays?: string;
     polylinePoints?: string;
     axisFormatMode?: string;
+    axisLabelFormating?: IntegerNumberFormattingOptions;
```
kuzmadom, line 24: Why only integer? There may also be float values on the axis.

src/ui/constants/visualizations/scatter.ts
```diff
@@ -61,6 +61,8 @@ export const SCATTER_VISUALIZATION: GraphShared['visualization'] = {
                 hideLabels: 'no',
                 labelsView: 'auto',
                 holidays: 'off',
+                axisLabelDateFormat: 'auto',
+                axisLabelFormating: {},
```
kuzmadom, line 65: Are we sure we need these values? It seems that the fields are optional

src/ui/libs/DatalensChartkit/modules/data-provider/charts/node.ts
```diff
@@ -364,6 +364,12 @@ function applyChartkitHandlers(args: {
                 ChartkitHandlersDict[ChartkitHandlers.WizardLabelFormatter];
         }
 
+        if (libraryConfig.xAxis?.labels?.formatter === ChartkitHandlers.WizardAxisFormatter) {
+            libraryConfig.xAxis.labels.formatter = ChartkitHandlersDict[
+                ChartkitHandlers.WizardAxisFormatter
+            ](libraryConfig.xAxis?.labels?.format);
```
kuzmadom, line 370: Only for the X-axis?

src/ui/libs/DatalensChartkit/modules/data-provider/charts/chartkit-handlers.ts
```diff
@@ -18,8 +19,16 @@ export const ChartkitHandlersDict = {
...
     [ChartkitHandlers.WizardTreemapTooltipFormatter]: wizardTreemapTooltipFormatter,
+    [ChartkitHandlers.WizardAxisFormatter]: wizardAxisFormatter,
 };
 
+function wizardAxisFormatter(format: string) {
+    return (point: Highcharts.AxisLabelsFormatterContextObject<number>) => {
+        const dateTimeValue = dateTimeUtc({input: point.value});
+        return dateTimeValue?.isValid() ? dateTimeValue.format(format) : point.value;
+    };
+}
```
kuzmadom, line 30: If the method is intended only for an axis with the datetime type, then I suggest reflecting this in its name.

src/ui/units/wizard/components/Dialogs/DialogPlaceholder/utils.ts
```diff
@@ -33,3 +39,20 @@ export function getAxisModeTooltipContent(reason: AxisModeDisabledReason) {
...
+    const isCastNumeric = isNumberField({data_type: formattingDataType});
+    const isDate = isDateField(field);
+    const isFormatless = !isCastNumeric && !isDate;
+
+    return {
+        formattingDataType,
+        isNumeric,
+        isCastNumeric,
+        isDate,
+        isFormatless,
```
kuzmadom, line 56: I don't think it's a good idea to do all the calculations if only a part is used in the end. This increases the preparation time for the chart config

src/ui/constants/visualizations/scatter.ts
kuzmadom: Then maybe it's worth fixing the chart type change instead of adding values here?

src/server/modes/charts/plugins/datalens/preparers/bar-x/highcharts.ts
```diff
@@ -134,21 +138,35 @@ export function prepareHighchartsBarX(args: PrepareFunctionArgs) {
...
+                    xPlaceholderSettings,
+                    x,
+                    xAxisType,
+                );
+                const isNotLogarithmic =
                     isXDiscrete &&
                     xPlaceholderSettings &&
-                    xPlaceholderSettings.type !== 'logarithmic'
-                ) {
+                    xPlaceholderSettings.type !== 'logarithmic';
```
kuzmadom, line 149: It is not entirely correct to add a discreteness check to the condition of the logarithmic axis. These should be two parallel conditions.

src/server/modes/charts/plugins/datalens/preparers/bar-x/highcharts.ts
```diff
@@ -134,21 +138,35 @@ export function prepareHighchartsBarX(args: PrepareFunctionArgs) {
...
-                            formatter: ChartkitHandlers.WizardXAxisFormatter,
-                        };
-                    }
                 } else if (xIsDate) {
                     customConfig.xAxis.type = 'datetime';
                 }
 
+                customConfig.xAxis.labels = {
+                    formatter,
+                    format,
```
kuzmadom, line 167: Does it really make sense to add both parameters?

src/server/modes/charts/plugins/datalens/types.ts
```diff
@@ -97,5 +97,5 @@ export type AxisOptions = {
         style?: React.CSSProperties;
         useHTML?: boolean;
     };
-    labels?: any[];
+    labels?: any;
```
kuzmadom, line 100: If you change any, it's better to make it more precise, rather than the other way around

PR #2809, opened 2025-08-22
Add StatusIllustration to a Navigation error layout

bttlvd-ya:


src/ui/components/Navigation/Core/ErrorView/ErrorView.scss
```diff
@@ -6,7 +6,7 @@
     flex-grow: 1;
     height: 100%;
 
-    &__message {
+    .g-placeholder-container__title {
```
mournfulCoroner, line 9: you should not use a private component classes unless absolutely necessary. maintainers do not guarantee their stability. Components are often transferred between libraries or redone. If something like this happens, it will be difficult to track this place

PR #2828, opened 2025-08-28
Add dataset revisions

dvortsovs:


src/ui/units/datasets/store/constants.ts
```diff
@@ -47,9 +47,12 @@ export const initialPreview: DatasetReduxState['preview'] = {
 };
 
 export const initialState: DatasetReduxState = {
+    isUpdatingDataset: false,
```
korvin89, line 50: Actually it is fetching (according to the place set this option in)

Btw, it called `isRefetchingDataset` directly at the place of use

src/ui/units/datasets/store/actions/creators/dataset.js
```diff
@@ -267,7 +285,7 @@ export function initialFetchDataset({datasetId}) {
     };
 }
 
-export function fetchDataset({datasetId}) {
+export function fetchDataset({datasetId, rev_id}) {
```
korvin89, line 288: It is a dead code, we will delete it later, there is no need to modify it

PR #2834, opened 2025-09-01
Add e2e test for dashboard selector with invalid field type

alexey-o-k:
- add markdown field `invalid_field` into dataset **Dataset** and create selector on this one stored in the **Selector test :: invalid field type** dashboard.
- build e2e test to check that this selector can be edit w/o errors


imsitnikov (approved):

tests/opensource-suites/dash/selectors/selectorWithInvalidField.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+            await expect(page.locator(slct('type-error'))).toHaveCount(0);
+
+            await dashboardPage.selectOptionByIndex(DialogControlQa.fieldSelect, SelectQa.LIST, 0);
+
+            await dashboardPage.controlActions.dialogControl.applyControlSettings();
+
+            await page.click(slct(COMMON_SELECTORS.ACTION_PANEL_CANCEL_BTN));
+
+            await page.waitForSelector(slct('dialog-control'), {state: 'detached'});
```
Marginy605, line 27: It's better use it from this constant https://github.com/datalens-tech/datalens-ui/blob/main/src/shared/constants/qa/control.ts#L27

tests/opensource-suites/dash/selectors/selectorWithInvalidField.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    datalensTest(
+        'Should successfully open edit modal when attempting to edit existing selector with incompatible field type',
+        async ({page}) => {
+            await openTestPage(page, DashUrls.DashboardWithInvalidSelector);
+            const dashboardPage = new DashboardPage({page});
+            dashboardPage.forceEnterEditMode();
+
+            await page.click(slct(DashKitOverlayMenuQa.SettingsButton));
+
+            await expect(page.locator(slct('type-error'))).toHaveCount(0);
```
Marginy605, line 19: Looks like this should be used from common constant for qa attrs 
example - https://github.com/datalens-tech/datalens-ui/blob/main/src/shared/constants/qa/control.ts#L1

tests/opensource-suites/dash/selectors/selectorWithInvalidField.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import DashboardPage from '../../../page-objects/dashboard/DashboardPage';
+import {DashKitOverlayMenuQa, DialogControlQa, SelectQa} from '../../../../src/shared';
+
+datalensTest.describe('Dashboards - Selector behavior with invalid field types', () => {
+    datalensTest(
+        'Should successfully open edit modal when attempting to edit existing selector with incompatible field type',
+        async ({page}) => {
+            await openTestPage(page, DashUrls.DashboardWithInvalidSelector);
+            const dashboardPage = new DashboardPage({page});
+            dashboardPage.forceEnterEditMode();
```
Marginy605, line 15: Bad way for future. Now you have created this dash only for your test for only one case. But it could be used some time later for other test or test case. So you may receive lot's of flaps for parallel tests runing. 
We use pattern that we copy dash in before section and deleting this copy in the after section. So that we have untouched dash without any chance of mutation or using it in parallel test case.

Marginy605 (changes requested):

tests/opensource-suites/dash/selectors/selectorWithInvalidField.test.ts
alexey-o-k: fixed

Marginy605 (approved):

PR #2905, opened 2025-09-18
Add dashboards to public api

stepanenkoxx:
1. Public api config refactored
2. Added implementation of different zod schemas for public api and gateway actions
3. Different swagger for each api version added
4. Api version moved from path to header
5. Scope and types filters added to delete entry actions
6. Dashboards sdk moved from server to shared
7. Dashboard actions and schemas added ( **It's not fully implemented yet, so you don't have to pay too much attention to these schemas and actions during the code review, they will still be finalized in next PR** )


src/shared/zod-schemas/dash.ts
```diff
@@ -0,0 +1,276 @@
...
+    dependentSelectors: z.boolean(),
+    globalParams: z.record(z.any(), z.any()).optional(),
+    hideTabs: z.boolean(),
+    hideDashTitle: z.boolean().optional(),
+    expandTOC: z.boolean(),
+    assistantEnabled: z.boolean().optional(),
+});
+
+// Data definition
+export const dataSchema = z.object({
```
Marginy605, line 259: Is it inner scheme? Or it will be for public?

stepanenkoxx: now its both inner and public, but we can make different schema for public

src/shared/zod-schemas/dash.ts
```diff
@@ -0,0 +1,276 @@
...
+    salt: z.string().min(1),
+    schemeVersion: z.literal(DASH_CURRENT_SCHEME_VERSION).default(DASH_CURRENT_SCHEME_VERSION),
+    tabs: z.array(tabSchema),
+    settings: settingsSchema,
+    supportDescription: z.string().optional(),
+    accessDescription: z.string().optional(),
+});
+
+// Main dashboard API validation schema
+export const dashSchema = z.object({
```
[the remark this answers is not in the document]
stepanenkoxx: it was like this in flops PR, I will review and fix all schemas in next PR

imsitnikov (approved):

PR #2929, opened 2025-09-25
Disable multiple dataset chart sort

dvortsovs:


src/ui/units/wizard/containers/Wizard/SectionVisualization/AddField/AddField.tsx
```diff
@@ -30,20 +30,31 @@ interface Props extends StateProps {
...
+            Boolean(disabledTextOrigin) ||
             addableFields.length === 0 ||
             (Number.isInteger(capacity) ? capacity! <= items.length : false);
-        const disabledText =
+        const disabledTextLocal =
             addableFields.length === 0
                 ? i18n('wizard', 'tooltip_no-available-fields')
                 : capacityError;
+
+        const disabledText = disabledTextOrigin || disabledTextLocal;
```
kuzmadom, line 57: the names are not very clear - local because it is defined in this component and not from above by a prop? I would expect the logic to gather in one place and not be distributed in different components.

kuzmadom (approved):

PR #2948, opened 2025-10-01
Fix invalid date when export xlsx or csv

dvortsovs:


src/server/modes/charts/plugins/datalens/preparers/flat-table/helpers/footer/index.ts
```diff
@@ -57,12 +57,18 @@ export const prepareFooterValue = (
...
-    const {column, value, columnValuesByColumn, colorsConfig, defaultColorPaletteId} = args;
+    const {column, columnIndex, value, columnValuesByColumn, colorsConfig, defaultColorPaletteId} =
+        args;
 
     const cell: TableCommonCell | BarTableCell = {
         value,
         css: TABLE_TOTALS_STYLES,
     };
+
+    if (columnIndex === 0) {
```
kuzmadom, line 68: If this implicit behavior is repeated in two places, then it is better to put it in a common utility function. In order not to miss an error when changing the logic

kuzmadom (approved):

kuzmadom (approved):

PR #3389, opened 2026-01-15
Add line width setting for wizard line charts

arsen-afaunov:


kuzmadom:
Screenshot 2026-01-16 at 17 19 23
If you enter the line width value 2000 for example. It would be necessary to limit

kuzmadom (changes requested):
The Line width parameter is incorrectly displayed and can be modified in the settings dialog for a Scatter-type chart. Since scatter plots typically use individual markers (points) and not connected lines, this setting should not be applicable and should be hidden

src/ui/units/wizard/components/Dialogs/DialogShapes/tabs/DialogShapesChartSettingsTab.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import {DialogLineWidth} from '../DialogLineWidth/DialogLineWidth';
+import type {ShapesState} from '../DialogShapes';
+
+type Props = {
+    shapesState: ShapesState;
+    onChartLineWidthChange: (nextLineWidth: string) => void;
+};
+
+export const DialogShapesChartSettingsTab: React.FC<Props> = ({
```
kuzmadom: It is better not to use React.FC<Props>

src/shared/types/config/wizard/v15.ts
```diff
@@ -311,6 +311,8 @@ export type V15ColorsConfig = {
 
 export type V15ShapesConfig = {
     mountedShapes?: Record<string, string>;
+    mountedShapesLineWidths?: Record<string, number>;
+    chartLineWidth?: number;
```
kuzmadom: Is it possible to refer to it without using a `chart`? The whole config is already related to the chart, so this doesn’t add any meaning.

